PR #9, opened 2018-11-28
added eor.py for rough sky-locked but noise-like eor signals

nkern:
More realistic EoR signals can be added to this module in the future. This is
modeled after the `diffuse_foreground` function in `foregrounds.py`.

This also creates `utils.py` for repo-wide functions, specifically access to fringe-rate filtering
functions. I added the ability in `rough_fringe_filter` to use a gaussian filter, in addition to a top-hat filter.

### DIFF
--- a/hera_sim/__init__.py
+++ b/hera_sim/__init__.py
@@ -6,5 +6,7 @@ import rfi
 import sigchain
 import vis
 import version
+import eor
+
 #import antpos
 __version__ = version.version
--- a/hera_sim/__init__.py
+++ b/hera_sim/__init__.py
@@ -7,6 +7,7 @@ import sigchain
 import vis
 import version
 import eor
+import utils
 
 #import antpos
 __version__ = version.version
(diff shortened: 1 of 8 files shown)


steven-murray:
Just as a note, I am of the opinion that variable names should be fully spelled out, unless there is very good reason not to do so. So variables like ``dly`` should be ``delay``. It just reduces a small amount of mental overhead when reading code, at almost zero cost to the time of actual coding (if using any kind of IDE). 

Sorry, this is probably not the best place to make this comment, but I thought I should somewhere.

nkern:
@steven-murray point well taken. If you traverse the HERA-Team stack you'll find common abbreviations for certain variables, like `freqs` for frequencies or `spws` for spectral windows (the latter being adopted in `CASA` and other software as well), which I personally think are fine abbreviations for particularly verbose names, but for variables that are already quite short we should probably expand them...

PR #10, opened 2018-11-28
added reflections.py for simulating reflections

nkern:
Adds functions for simulating and inserting cable reflections and feed-to-feed reflections into visibility data. This is different than `hera_cal.reflections.py` which is responsible for reflection modeling and removal.

This is also different than `hera_sim.sigchain.gen_xtalk`, which simulates random, time-independent crosstalk. This module simulates time-dependent feed-to-feed crosstalk.

### DIFF
--- a/hera_sim/reflections.py
+++ b/hera_sim/reflections.py
@@ -1,0 +1,82 @@
+"""A module for generating reflections"""
+
+import numpy as np
+import aipy
+
+def auto_reflection(vis, freqs, amp, dly, phs, conj=False):
+    """
+    Insert an auto reflection into vis data. An auto reflection
+    is one that couples an antenna's voltage stream with itself.
+    Examples include cable reflections and dish-to-feed reflections.
+
+    Args:
+        vis : 2D complex ndarray, shape=(Ntimes, Nfreqs)
+        freqs : 1D ndarray, holds frequencies in GHz
+        amp : float, reflection amplitude
+        dly : float, reflection delay [nanosec]
+        phs : float, reflection phase [radian]
+        conj : bool, if True, conjugate reflection coefficient
+
+    Returns:
+        out_vis : contains input vis with the auto reflection
+    """
+    # form reflection coefficient
+    eps = amp * np.exp(2j * np.pi * freqs * dly + 1j * phs)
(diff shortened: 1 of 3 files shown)


nkern:
screen shot 2018-11-28 at 11 30 55 am
^Simulating Feed-to-Feed Reflection Crosstalk

screen shot 2018-11-28 at 11 32 14 am
^Simulating Cable Refections

hera_sim/reflections.py
```diff
@@ -0,0 +1,82 @@
...
+        eps = eps.conj()
+
+    # multiply into data
+    out_vis = vis * (1 + eps)
+
+    return out_vis
+
+
+def cross_reflection(vis, freqs, autocorr, amp, dly, phs, conj=False):
+    """
```
anguta: From what is in this documentaiton, its confusing as to why this situation is distinct from the ordinary reflection above -- if the source of the cross reflection is coming from the sky --. I suspect that you are referring to reflections of non-sky-signals such as antenna thermal noise? If so, I think you should say this explicitly in the documentation.

hera_sim/reflections.py
```diff
@@ -0,0 +1,82 @@
...
+    Returns:
+        out_vis : contains input vis with the auto reflection
+    """
+    # form reflection coefficient
+    eps = amp * np.exp(2j * np.pi * freqs * dly + 1j * phs)
+    if conj:
+        eps = eps.conj()
+
+    # multiply into data
+    out_vis = vis * (1 + eps)
```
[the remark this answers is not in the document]
nkern: hmm.. good point, i'm not entirely sure about how harmonics above the max delay of the data would show up, I also suspect they wouldn't appear. In anycase I'll probably punt on that until I see a clear need for it, for the time being one can add in the harmonics themselves via repeated calls of this function.

hera_sim/reflections.py
nkern: I added some math to make it more clear hopefully. I do mean reflections of "signal" (any kind of signal, noise or sky or anything else).

anguta: Thanks for clarifying this!

PR #20, opened 2019-01-09
Simulator

steven-murray:
I believe that much more can be done on this before merging.

So far, the class has methods for adding all effects that exist within ``hera_sim``, which can be specified by str name or callable function, plus any other compatible user-defined function. Any time one of these is called, the underlying ``UVData`` object is updated, both its data and its history. There are also methods for reading/writing the object. I also added a jupyter notebook showing the basic usage. 

Some ideas on what could be done, for discussion:

**THINGS THAT COULD BE BETTER**
- Every component model should have similar kinds of behaviour, in terms of how it adds to the visibilities. **[EDIT: See comments below on homogenization of various functions]**
- Would be nice to be able to "register" component models, which should automatically perform a check of whether that function is compatible with the required format, and will enable interactively identifying what models are available. Possibly even accessing the docstring of each model from the main class. **[EDIT: I think this can be re-assessed once the PR is merged and we've used it for a while.]**
- There could be some kind of check which ensures proper ordering of effects -- some effects do nothing
  if the visibilities are zero prior to the effects being added. I don't know the details, but it's possible
  that the effects should be administered in a definite prescribed order, in which case this should be done
  automatically from an orderless list passed by the user. If the order is only semi-important, it would
  be enough to have checks before running each, and a warning. **[EDIT: this has been partially fixed in this PR by emitting warnings if relative effects are added before absolute ones]**
- Could potentially add antpos models to the initialiser to create arrays, in addition to manual specification. **[EDIT: I think this can be done separately from this PR, and I've created an issue for it]**
- There are a lot of attributes of HERA lying around the codebase (eg. HERA_LOT_LON_ALT). I think it makes more sense to compile these kinds of attributes into a config file (YAML? PY?) which can be accessed directly. All functions in hera_sim should *by default* use this HERA config file, but can be set to another file (and each parameter can be individually overwritten in the function call). I'm assuming there's already a kind of file format for this kind of thing, but if not we should make it. **[EDIT: this is separate to this PR so I am moving it to it's own issue]**
- resample_Tsky seems like a bad name. **[EDIT: again, not really part of this PR]**
- The sigchain.py module follows a different kind of format from other modules, such that several functions
  have to be called in chain to add a single effect, which means for now that it is implemented differently
  in Simulator(). This could easily be homogenized. **[EDIT: I think the homogenization of model functions can be improved a lot, but can be done after this PR is merged. Will make an issue]**

**QUESTIONS**
- The vis.py module seems to do nothing useful at this point (and is defined incorrectly at several points)... is that true? **[EDIT: it does something useful, but is very out-dated. It needs to be re-written to use ``pyuvdata``. Created an issue]**
- Would it make sense to escalate the ``empty_uvdata`` function to pyuvdata itself? It seems like a generally useful method. We could over-ride it in hera_sim to provide HERA defaults. **[EDIT: not really related to this PR]**
- Would it be good to keep a reference to the array that's *added* by each component? Or just blindly accumulate them? Or both? **[EDIT: This has been addressed in the PR]**
- Is the 'autocorr' argument in cross_reflections just the output of 'auto_reflections'? If so, is auto_reflections not meant to be added directly, just cross_reflections? **[EDIT: Addressed in comments below]**


hera_sim/simulate.py
```diff
@@ -0,0 +1,335 @@
...
+    """
+    Primary interface object for hera_sim.
+
+    Produces visibility simulations with various independent sky- and instrumental-effects, and offers the resulting
+    visibilities in :class:`UVData` format
+    """
+
+    def __init__(
+        self,
+        data_filename=None,
```
nkern: can we add a keyword argument that "whitens" the input data if provided. a use-case is that we have a `UVData` that we want to use as a template to get metadata, but we want to start over from scratch. So something like
```
uv.data_array[:] = 0.0
uv.flag_array[:] = False
uv.nsample_array[:] = 1.0
```
if we want to whiten it out...

nkern: Hmm, interesting idea, but no the metadata are intimately tied to the data array, so I think whitening the data array should be accompanied by also whitening the flags and nsamples. If in the future we develop some need for it we can always add that in later--but for now better to not over-optimize I think.

hera_sim/simulate.py
```diff
@@ -0,0 +1,335 @@
...
+                file_type = filename.split(".")[-1]
+            except IndexError:
+                raise ValueError(
+                    "if no file_type is given, the filename must be suffixed with either .miriad, .uvfits or .uvh5"
+                )
+
+        if file_type not in ["miriad", "uvfits", "uvh5"]:
+            raise ValueError("file_type must be one of 'miriad', 'uvfits' or 'uvh5'")
+
+        getattr(self.data, "write_%s" % filename.split(".")[-1])(filename, **kwargs)
```
nkern: looks like this should actually be
```
getattr(self.data, "write_%s" % filetype)(filename, **kwargs)
```
This should have failed but you got lucky b/c your test was using a `uvh5` file whose suffix happens to also be the file type. This isn't true for `miriad`.

hera_sim/simulate.py
```diff
@@ -0,0 +1,335 @@
...
+
+    def _iterate_baselines(self, with_conj=True):
```
nkern: This function iterates over all possible baselines based on `antenna_numbers`, which might include a lot of baselines that don't actually exist in the object for a number of reasons. I think we can solve two problems at once by making this something like
```
def _iterate_antpairpols(self):
  for app in self.data.get_antpairpols()
    blt_inds = self.data.antpair2ind(app)
    pol_ind = self.data.get_pols().index(app[2])
    yield app[0], app[1], app[2], blt_inds, pol_ind
```
Once we do that, in each `add_*` function we can use the relevant pol_ind in the `self.data.data_array[:, :, :, pol_ind]]` call, does that make sense?

hera_sim/simulate.py
```diff
@@ -0,0 +1,335 @@
...
+                lsts=lsts,
+                fqs=self.data.freq_array[0]
+                * 1e-9,  # Axis 0 is spectral windows, of which at this point there are always 1.
+                bl_len_ns=bl_len_m * 1e9 / 3e8,
+                **kwargs
+            )
+
+            self.data.data_array[
+                ind, 0, :, 0
+            ] += vis  # TODO: not sure about only using first pol.
```
nkern: See comment above for how to use the proper polarization index. At some point we need to expose a `polarization` kwarg in `model`, but let's worry about that once we have models that actually take a polarization..

hera_sim/simulate.py
```diff
@@ -0,0 +1,335 @@
...
+                telescope_location=telescope_location,
+                telescope_lat_lon_alt=telescope_lat_lon_alt,
+                object_name=object_name,
+                start_jd=start_jd,
+                vis_units=vis_units,
+            )
+
+        else:
+            self.data = self._read_data(self.data_filename, pols, **kwargs)
+
```
nkern: Let's add an antpos dictionary to `self` via something like
```
antpos, ants = self.data.get_ENU_antpos()
self.antpos = dict(zip(ants, antpos))
```
using whatever variables you like

hera_sim/simulate.py
```diff
@@ -0,0 +1,335 @@
...
+        Add an EoR-like model to the visibilities.
+
+        Args:
+            model (str or callable): either a string name of a model function existing in :mod:`~hera_sim.eor`, or
+                a callable which has the signature ``fnc(lsts, fqs, bl_len_ns, **kwargs)``.
+            **kwargs: keyword arguments sent to the EoR model function, other than `lsts`, `fqs` and `bl_len_ns`.
+        """
+        for ant1, ant2, ind in self._iterate_baselines():
+            lsts = self.data.lst_array[ind]
+            bl_len_m = self.data.uvw_array[ind][
```
nkern: Lets use the `self.antpos` dictionary instead of this, b/c `uvw_array` can change depending on the phasing of the data...

nkern: Or maybe `uvw_array` is the correct attribute to use, to account for what phased visibilities actually see on the sky.. not exactly sure about this, and I think it may be dependent on the assumptions of the `model` function. Better to use the `antpos` in the mean time and we can make this question an issue and come back to it later.

hera_sim/simulate.py
```diff
@@ -0,0 +1,335 @@
...
+        channel_bw=0.1 / 1024.0,
+        instrument="hera_sim",
+        telescope_location=None,
+        telescope_lat_lon_alt=None,
+        object_name="sim_data",
+        start_jd=2458119.5,
+        vis_units="uncalib",
+        **kwargs
+    ):
+
```
nkern: the `__init__` needs a docstring

hera_sim/simulate.py
```diff
@@ -0,0 +1,335 @@
...
+                telescope_lat_lon_alt=telescope_lat_lon_alt,
+                object_name=object_name,
+                start_jd=start_jd,
+                vis_units=vis_units,
+            )
+
+        else:
+            self.data = self._read_data(self.data_filename, pols, **kwargs)
+
+    def _read_data(self, filename, polarizations, **kwargs):
```
nkern: why is `polarizations` special? It is a `kwarg` of `uv.read` so you can just lump it in with `**kwargs`

hera_sim/simulate.py
nkern: Let's also assert for now that the data are drift phased, b/c I think many of our functions assume drift-phased data:
something like
```
assert self.data.phase_type == 'drift', "uvd.phase_type must be 'drift'"
```

nkern:
Also, the `autocorr` intpu to `cross_reflections` is the autocorrelation visibility. It doesn't distinguish currently between `ant1, ant1` or `ant2, ant2`, so probably for now its okay to just use feed it `self.data.get_data(ant1, ant1, pol)` or something

hera_sim/simulate.py
steven-murray: Thanks for picking this up. I've changed this in 47c08ac73ed6c8ffe9243b5727ae57209bdb01dd. Now, if the user passes a file with suffix ".uv" it will automatically interpret as miriad, whereas .uvfits and .uvh5 will obviously by uvfits or uvh5. And it will actually use the assigned file_type, rather than the suffix itself.

hera_sim/simulate.py
steven-murray: Yeah, I really was just taking a stab at it. This looks like a good addition. I've added it in 5e528d5a9505e6d5df19ae9fd57c424265a3d5c4

hera_sim/simulate.py
steven-murray: This is also fixed (for now) in 5e528d5a9505e6d5df19ae9fd57c424265a3d5c4

hera_sim/simulate.py
steven-murray: What would be the purpose of this? I mean, it exposes the antpos a little more directly than accessing through the data object, but where is this used in the Simulator?

steven-murray: Ah, I see, is this just attached to the following comment?

hera_sim/simulate.py
steven-murray: I don't really understand the subtleties of this. It sounds like it's kind of ambiguous at the moment. If so, maybe it's better to leave it as-is, and once merged, create an issue for it where it can be hashed out in full detail.

hera_sim/simulate.py
steven-murray: I added this check in 5cb53d206c875fcf143bd5a1f89e7c7b60ffd4a0. I also created a framework for such compatibility checks, and a new Exception class, "CompatibilityException" which will be raised if any of these conditions fail.

steven-murray:
@nkern concerning the autocorrelation, I'm not _quite_ sure what you mean here. Could you do a quick commit for just that change?

hera_sim/simulate.py
nkern: I think its more safe to use `antpos` over `uvw_array`, b/c `uvw_array` can definitely be changed depending on the phasing of the data, so I think we should default to that for now, and I've created an issue for it in the meantime.. LMK if you have any thoughts why we should default to `uvw_array` over `antpos` for now

hera_sim/simulate.py
nkern: nice!

hera_sim/simulate.py
nkern: Hmm, an interesting idea, but I would be a little hesitant to automatically infer the filetype from the suffix, b/c it is in principle completely up to the user to define this--and someone could call a `uvfits` file something like`uvf`. Furthermore, a `uvh5` file is just an `HDF5` file which could be called `.h5` or `.hdf5`, and additionally we often have miriad files that *dont* have the suffix `uv`, but something more complicated, like `uvOCSL` or `uvRX` etc.

My opinion would be to keep the `filetype` kwarg to determine which `read_` func to call, which should default to `uvh5`. Perhaps @jaguirre has thoughts on this too...

hera_sim/simulate.py
nkern: Yup

hera_sim/simulate.py
plaplant: I don't mean to butt in too much, but I saw this comment and thought I'd say that we've done something like this in the generic read function in pyuvdata: https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/29d4a0802ea24a46b551cc55624bd332fbcec563/pyuvdata/uvdata.py#L2331

Basically, we test if the "file" is actually a directory (which means it's a miriad dataset or measurement set), and look for a `vartable` (miriad) or `OBSERVATION` (measurement set) file in the directory. For uvfits and uvh5, we look at the file suffix, but you could imagine doing something more sophisticated like doing a try-except block where you open and access the file using h5py or astropy.io.fits. I know h5py checks the file's "magic number", and can immediately tell if it's an HDF5 file or not.

Mostly just wanted to say that there's precedent for inferring file type from the name, but it's obviously not foolproof :)

steven-murray: OK, I've done the easy thing and now it just ignores the file extension altogether, and uses file_type instead (which defaults to uvh5). I think this is just easier and less error prone.

hera_sim/simulate.py
steven-murray: @jaguirre  agrees. I've changed this in b5a1c00865d817f67b946b2e2a45d233c3961f88.

PR #21, opened 2019-01-11
Better foregrounds

steven-murray:
Fix some issues with current implementation of foregrounds

Stuff to do:

- pnt source fg still don't take into account baseline orientation


hera_sim/foregrounds.py
```diff
@@ -12,41 +13,61 @@ def diffuse_foreground(Tsky, lsts, fqs, bl_len_ns, bm_poly=noise.HERA_BEAM_POLY,
...
+
+    # Fringe-rate max could be zero if bl_len = 0 (eg. auto-correlations)
+    if fr_max > 0:
+        # If not zero, do a fringe rate filter
+        dt = 1.0 / (fr_max_mult * fr_max)  # over-resolve by fr_mult factor
+        ntimes = int(np.around(aipy.const.sidereal_day / dt))
+        lst_grid = np.linspace(0, 2 * np.pi, ntimes, endpoint=False)
+        nos = Tsky(lst_grid, fqs) * noise.white_noise((ntimes, fqs.size))
+
+        nos, ff, frs = utils.rough_fringe_filter(nos, lst_grid, fqs, bl_len_ns, fr_width=fr_width)
```
adampbeardsley: Since this is a wide-range PR... I think we should make use of fr_width, probably by default. We should at least figure out a reasonable value and include it in the doc string, but possibly just automatically use it.

adampbeardsley: On a separate note, I think this filter gives us the opposite behavior than we want for diffuse structure. We start with Tsky power, but then filter it -- the shorter the baseline, the more power we filter out. But the diffuse power should have _more_ power at shorter baselines.

PR #27, opened 2019-02-13
Edit to vis.vis_cpu to re-introduce freq factor that somehow got dele…

AaronParsons:
…ted several months ago and committed.


steven-murray:
This fails because of an error in hera_cal, which really just doesn't have to be installed anyway. If you remove the hera_cal line from the .travis.yml, then it should test OK, and we can merge.

jsdillon:
Why is hera_cal erroring? We're passing our own tests.

steven-murray:
@jsdillon it is failing when attempting to install hera_cal, complaining that sklearn is not installed. We could just install it before hera_cal, of course, but AFAICT hera_sim does not need hera_cal anyway (except for in the notebook tutorial), so we could probably just not have it in the .travis.yml install process.

steven-murray:
We've decided that even though hera_cal is not currently used, it will be soon, so we should just add scikit-learn to the .travis.yml installation process.

steven-murray:
I will make this fix in the Simulator branch, so don't worry about it here. Let's just pull this in.

PR #29, opened 2019-02-16
moved reflections to sigchain

nkern:
Moved code in `reflections.py` over to `sigchain.py` as we noted a few telecons ago. Also updated the nomenclature of the functions as I've changed them in my memo since we talked last about them. Lastly, updated the simulator `add_` functions.


nkern:
@steven-murray I need some of your help updating the `add_xtalk` and `add_gains` methods in the `Simulator` class. In trying to update them with new models, I realized that the structure we have now is a little too rigid: we need a way to pass arguments (not kwargs) to the `add_` methods, which will themselves do a *little* bit of preprocessing before iterating over the module level component functions. For the time being, I've commented out the `@_model` decorator b/c I can't quite parse what you've done to it so that it only takes an optional `model` argument. Can you allow it to take normal arguments that are passed to the `add_` method namespace so we can access them like I am doing now?

Feel free to commit to this branch if you have a quick fix, thanks!

steven-murray:
The issue with ``@_model`` decorator is indeed slightly tricky. Let me try to summarize what the difficulty is. First, I don't think there should be any trouble with having an argument that is not a kwarg. It should get passed through the decorator correctly, and won't be passed to the model as a kwarg. So you can do whatever logic you like before that.

The bigger problem is that the structure expects the following:

1. Any model functions (i.e. those found in other modules) are single-call functions, i.e. you don't need to call a chain of functions to get any given effect.
2. These functions should have some arguments which are _fully defined_ by the UVData object (so arguments such as freq, baseline length etc.), and then some arguments which are specific to that model. 
3. Models that are different models for the same effect should have the same signature in terms of those parameters that are fully defined by the UVData object, but can have arbitrary other parameters.

This combination of expectations enables the ``@_model`` decorator to work, so that any given ``add_xxx`` method can be supplied with a specific model, but the called signature can remain the same in terms of what the class itself has to calculate (and then the user can pass in whatever other arguments that particular model takes). 

In your case, this doesn't work very well, because it does not conform to expectation 1 or 3. That is, you need to call _two_ functions to actually apply the xtalk to the visibilities, and the first of these can be called in different ways in terms of its class-defined arguments (specifically, ``whitenoise`` requires only ``freq``, while ``cross_coupling`` requires ``freq`` _and_ ``autovis``). 

The first of these problems actually really isn't a problem in this case because the second function that needs to be called is very simple and will be the same for any kind of model passed, so it can be hard-wired in. 

The second of these problems is more difficult. One solution (which is the one I previously took) would be to split the model up into two different ``add_xxx`` methods. That doesn't seem to be particularly nice because it would seem that these models should be naturally grouped as the same effect. Another would be to leave off the ``@_model`` decorator altogether, as you've done. This also isn't particularly nice because it breaks the homogeneity of the class, and also means that the ``history`` of the UVData object isn't automatically written etc.

Another solution would be to specify that all xtalk models should be able to take the autovis argument, which they can then ignore. 

I would prefer either my solution (split it into two) or this last option, depending on how naturally grouped these models are. Are they different implementations/models of the same underlying kind of effect? Then go with the latter option (or something similar). Otherwise, they should be added as different kinds of effects to the Simulator, by specifying them as different ``add_xxx`` methods.

nkern:
@steven-murray it doesn't look like the `@_model` decorator accepts arguments other than the `model`. Here is the traceback I get when I try to feed it a single argument other than `model`
```
TypeError                                 Traceback (most recent call last)
<ipython-input-4-749ba32b93e1> in <module>()
----> 1 sim.add_sigchain_reflections([0])

/Users/nkern/Software/anaconda2/lib/python2.7/site-packages/hera_sim-0.0.1-py2.7.egg/hera_sim/simulate.pyc in new_func(obj, *args, **kwargs)
     94                 # For cases where there is no choice of model.
     95                 method = ""
---> 96                 func(obj, **kwargs)
     97 
     98             if add_vis:

TypeError: add_sigchain_reflections() takes exactly 2 arguments (1 given)

```
where i've defined it the class method as
```
    @_model(multiplicative=True)
    def add_sigchain_reflections(self, ants, **kwargs):
```
and am calling it as
```
    sim.add_sigchain_reflections([0])
```

LMK if you have any thoughts

steven-murray:
Oh excellent, I don't know how I missed that! That looks good. My only remaining concern would be that the ``mode`` parameter in ``add_xtalk`` should really just be the standard ``model`` -- all it is doing is selecting a model to use. The actual model that is passed through the decorator is a function, so you should be able to access its ``name`` attribute to do the logic you want. This keeps things a bit more consistent. In fact, I might take a stab at doing this, and you can let me know if that works for you.

steven-murray:
Check out the updated code to see if you like it. I just changed mode to model, and gave defaults to ants and bls (default all ants and all bls). 

If you strongly disagree with having model rather than mode, LMK.

nkern:
@steven-murray this looks fine to me! I just merged it w/ master, so whenever it passes travis feel free to approve and merge the branch! thanks!

steven-murray:
Thanks @nkern.

PR #30, opened 2019-02-21
general updates

nkern:
Mostly updates to functions in `eor`, `foregrounds` and `utils` and some smaller updates to other modules


hera_sim/foregrounds.py
```diff
@@ -7,61 +7,48 @@
...
-from scipy.interpolate import RectBivariateSpline
 
 from . import noise
 from . import utils
 
 
-def diffuse_foreground(Tsky, lsts, fqs, bl_len_ns, bm_poly=noise.HERA_BEAM_POLY, scalar=30.,
-                       fr_width=None, fr_max_mult=2.0):
-
+def diffuse_foreground(Tsky_mdl, lsts, fqs, bl_vec, bm_poly=noise.HERA_BEAM_POLY,
```
steven-murray: If you set Tsky_mdl not as the first parameter, it will make it possible to call it from the simulator class with Tsky_mdl as a kwarg, rather than requiring it as another parameter.

hera_sim/foregrounds.py
```diff
@@ -7,61 +7,48 @@
...
+    data = np.asarray(Tsky * 1e3 / noise.jy2T(fqs, bm_poly=bm_poly), np.complex)
 
-    fr_max = max(fr_max, fr_max_beam)
+    # multiply by white noise
+    data *= noise.white_noise((len(lsts), len(fqs)))
 
-    dt = 1.0 / (fr_max_mult * fr_max)  # over-resolve by fr_mult factor
-    ntimes = int(np.around(aipy.const.sidereal_day / dt))
+    # fringe rate filter across time using projected East-West distance
+    data, fringe_filter = utils.rough_fringe_filter(data, lsts, fqs, np.abs(bl_vec[0]), filter_type=fringe_filter_type, **fringe_filter_kwargs)
```
steven-murray: will this work OK if abs(bl_vec[0]) = 0?

hera_sim/foregrounds.py
```diff
@@ -7,61 +7,48 @@
...
 
-    lst_grid = np.linspace(0, 2 * np.pi, ntimes, endpoint=False)
-    nos = Tsky(lst_grid, fqs) * noise.white_noise((ntimes, fqs.size))
+    # delay filter across freq
+    data, delay_filter = utils.rough_delay_filter(data, fqs, np.linalg.norm(bl_vec), standoff=standoff, filter_type=delay_filter_type)
```
steven-murray: will this work for auto-correlations?

hera_sim/noise.py
```diff
@@ -7,23 +7,23 @@
...
+lsts = npz['lsts'] / 12. * np.pi
+lsts = np.concatenate([lsts[-10:]-2*np.pi, lsts, lsts[:10]+2*np.pi])
+HERA_Tsky_xx = npz['HERA_Tsky'][0].T
+HERA_Tsky_yy = npz['HERA_Tsky'][1].T
 HERA_Tsky_xx = np.concatenate([HERA_Tsky_xx[-10:], HERA_Tsky_xx, HERA_Tsky_xx[:10]])
 HERA_Tsky_yy = np.concatenate([HERA_Tsky_yy[-10:], HERA_Tsky_yy, HERA_Tsky_yy[:10]])
 HERA_Tsky_mdl = {}
-HERA_Tsky_mdl["xx"] = RectBivariateSpline(lsts, fqs, HERA_Tsky_xx)
-HERA_Tsky_mdl["yy"] = RectBivariateSpline(lsts, fqs, HERA_Tsky_yy)
+HERA_Tsky_mdl['xx'] = RectBivariateSpline(lsts, fqs, HERA_Tsky_xx, kx=4, ky=4)
```
steven-murray: are we certain that no extrapolation is being performed? otherwise it's much better to use kx,ky=1 usually

hera_sim/simulate.py
```diff
@@ -256,25 +256,21 @@ def add_eor(self, model, **kwargs):
...
-            bl_len_m = np.abs(self.antpos[ant1][0] - self.antpos[ant2][0])  # E-W baseline length
-
-            self.data.data_array[blt_ind, 0, :, pol_ind] += model(
-                lsts=lsts,
-                # Axis 0 is spectral windows, of which at this point there are always 1.
-                fqs=self.data.freq_array[0] * 1e-9,
-                bl_len_ns=bl_len_m * 1e9 / 3e8,
-                **kwargs
-            )
+            bl_vec = (self.antpos[ant1] - self.antpos[ant2]) * 1e9 / 2.99e8
```
steven-murray: we should probably start making a habit of using astropy constants rather than just hardcoding 2.99e8 etc.

hera_sim/simulate.py
```diff
@@ -283,24 +279,29 @@ def add_foregrounds(self, model, **kwargs):
...
             add_vis (bool, optional): whether to add the calculated visibilities to the underlying data array.
                 Default True.
-            **kwargs: keyword arguments sent to the foregournd model function, other than `lsts`, `fqs` and `bl_len_ns`.
+            **kwargs: keyword arguments sent to the foregournd model function, other than `lsts`, `fqs` and `bl_vec`.
         """
+        # frequencies come from zeroths spectral window
+        fqs = self.data.freq_array[0] * 1e-9
+
+        # prep kwargs
+        if model.__name__ == 'diffuse_foreground':
```
steven-murray: this wouldn't be necessary if either you 1) use keyword args when passing to the model function, or 2) specify the ``diffuse_foregrounds`` function with ``Tsky_mdl`` not as the first argument.

hera_sim/tests/test_simulator.py
```diff
@@ -115,11 +115,10 @@ def test_wrong_func():
     sim.add_eor("noiselike_EOR")  # wrong function name
 
 
-@raises(TypeError)
+@raises(AssertionError)
 def test_wrong_arguments():
     sim = create_sim()
-    sim.add_foregrounds(diffuse_foreground, Tsky_mdl=HERA_Tsky_mdl['xx'])
```
steven-murray: this test should actually fail with the original TypeError if my other comments on this function are incorporated.

steven-murray:
It seems that two reflections tests are also failing, so we should probably also fix those :-)

hera_sim/foregrounds.py
nkern: hmm, not sure how that would work, b/c `kwargs` is an unordered dictionary... if you can paste a code snippet below of what you mean it might help enlighten me

hera_sim/foregrounds.py
nkern: I believe so

hera_sim/foregrounds.py
nkern: I don't think so, because the visibility will still be complex. the auto-correlation should actually just be `Tsky` in this case..

hera_sim/simulate.py
nkern: yup

hera_sim/simulate.py
nkern: I think this is redundant with the comment you made above. I still don't see what you mean because 1) I am passing keyword args to the model function.

Do you mean that this would work if I define `Tsky_mdl` to be a kwarg in the original function `diffuse_foreground()`? That I think would work, but b/c `Tsky_mdl` is a required parameter for that function, it seems counterintuitive to not make it an argument

hera_sim/noise.py
nkern: It may be possible to extrapolate it and get asymptotic results, but the high-order interpolation serves an important purpose, which is to suppress high fringe-rate structure in the valid LST regimes. In the case that one gets wild results from bad extrapolation it will probably be apparent. Not sure if `RectBivariateSpline` gives extrapolation warnings, but `interp1d` does, however, `RectBivariateSpline` is supposed to be faster than `interp1d`, so not sure what the tradeoff is. I'm fine w/ it as is for now..

nkern:
@steven-murray okay I think this is in a spot where i'm good to merge it. lmk if you have any remaining comments on this, thanks!

hera_sim/foregrounds.py
steven-murray: If the signature of ``diffuse_foreground`` is ~

```python
def diffuse_foreground(lsts, fqs, bl_vec, Tsky_mdl=None, bm_poly=noise.HERA_BEAM_POLY,...):
    if Tsky_mdl is None:
        raise TypeError("Tsky_mdl is a required parameter of diffuse_foreground")
    ...
```

then in the ``Simulator`` class, the ``add_foreground`` function can leave off the ``Tsky_mdl`` argument explicitly, i.e.

```python
@_model()
    def add_foregrounds(self, model, **kwargs):
        for ant1, ant2, pol, blt_ind, pol_ind in self._iterate_antpair_pols():
            lsts = self.data.lst_array[blt_ind]
            bl_len_m = np.abs(self.antpos[ant1][0] - self.antpos[ant2][0])  # E-W baseline length

            self.data.data_array[blt_ind, 0, :, pol_ind] += model(
                lsts=lsts,
                fqs=self.data.freq_array[0] * 1e-9,
                bl_len_ns=bl_len_m * 1e9 / 3e8,
                **kwargs
            )
```
I'm not _absolutely_ convinced this is better than the way you've done it. It just seems more consistent with how the rest of the ``add_`` functions are implemented.

hera_sim/noise.py
steven-murray: Sounds good. Can we add a short comment there to this effect, to warn our future selves?

hera_sim/simulate.py
steven-murray: Yeah, this is redundant with the previous comment. I agree that there is a bit of ambiguity here as to the best way to do it (obviously your way works just as well as the one I've mentioned above). Having thought a little bit more about it, I think the reason I like the way I described above slightly more is because of encapsulation -- the API function doesn't contain information about the specific model its calling -- all of the information is contained within the model function itself. The model itself raises the error if it hasn't been called correctly. This means we can write arbitrarily many foreground model functions and not have to change the ``Simulator`` class. It also means that users can write their own foreground model functions and plug-n-play the same way without touching the main code. 

On the other hand, I agree that making ``Tsky_mdl`` a keyword parameter is a bit counter-intuitive. It could be made a required argument even in the case I outlined above, of course, at the expense of less-informative errors when it isn't passed correctly.

steven-murray:
@nkern yes it looks basically ready to merge. I would like to figure out the details of the foreground function, if you don't mind, if only because I think it sets a pattern for how we should approach this for other functions in the future. I'm happy to merge as-is if you maintain that modifying the Simulator class internally is the favored approach. Otherwise, I will merge after that one small edit.

hera_sim/foregrounds.py
nkern: Sure thing

PR #31, opened 2019-02-21
Documentation update

AaronParsons:
Completely backwards-compatible tweaks to docstrings and occasionally var names.  No behavior changes.

### DIFF
--- a/hera_sim/noise.py
+++ b/hera_sim/noise.py
@@ -136,7 +136,7 @@ def sky_noise_jy(Tsky, fqs, lsts, bm_poly=HERA_BEAM_POLY, B=None, inttime=10.7):
     # XXX fix below when jy2T changed to Jy/K
     T2jy = 1e3 / jy2T(fqs, bm_poly=bm_poly)  # K to Jy conversion
     T2jy.shape = (1, -1)
-    Vnoise_jy = T2jy * Tsky / np.sqrt(inttime * B) # see noise_study.py for discussion of why no factor of 2 here
+    Vnoise_jy = T2jy * Tsky / np.sqrt(inttime * B_Hz) # see noise_study.py for discussion of why no factor of 2 here
     return white_noise(Vnoise_jy.shape) * Vnoise_jy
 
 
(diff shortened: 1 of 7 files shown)


hera_sim/noise.py
```diff
@@ -25,40 +25,42 @@
...
 
-# XXX replicated from hera_pspec.pspec
 HERA_BEAM_POLY = np.array([8.07774113e+08, -1.02194430e+09,
                            5.59397878e+08, -1.72970713e+08, 3.30317669e+07, -3.98798031e+06,
                            2.97189690e+05, -1.24980700e+04, 2.27220000e+02])  # See HERA Memo #27
 
 
-# XXX replicated from hera_pspec.pspec
-def jy2T(f, bm_poly=HERA_BEAM_POLY):
+def jy2T(fq, bm_poly=HERA_BEAM_POLY):
```
steven-murray: should be fqs?

hera_sim/noise.py
```diff
@@ -67,22 +69,29 @@ def white_noise(size=1):
     )
 
 
+# XXX reverse fqs and lsts in this function?
 def resample_Tsky(fqs, lsts, Tsky_mdl=None, Tsky=180.0, mfreq=0.18, index=-2.5):
     """
-    Evaluate sky temperature as a function of LST and frequency.
+    Re-sample a model of the sky temperature at particular freqs and lsts.
```
steven-murray: it is not clear to me what "resample" means here -- to me it indicates statistical sampling, but this function seems to only evaluate the sky temperature?

hera_sim/noise.py
AaronParsons: just made that change.

steven-murray:
According to https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html the Google style docstrings have arguments specified as 

```python
name (type): description
    Optional extra description lines...
    
    And paragraphs...
```

It looks like your docstrings are almost fine, but you have the types after the colon, rather than in parentheses before the colon. 

It is useful to stick with the standard format so that Sphinx can parse it for the HTML documentation. We could of course change to numpy format if you prefer it, but then we have to change all docstrings.

hera_sim/noise.py
AaronParsons: Tsky_mdl is an interpolator, so you are resampling one set of data onto a new fqs/lsts grid

AaronParsons:
@steven-murray After these tests pass, should be good to go.

steven-murray:
Thanks @AaronParsons . I'll let you delete the branch if you want to.

PR #33, opened 2019-03-01
Visibility simulators

### LINKED ISSUE #26: vis.py
At the moment, the vis.py module is a bit all over the place. The only real functionality in it that really belongs there is the ``vis_cpu`` function, which is a quick visibility simulator (i.e. from a sky-locked signal). It's probably a good idea to rewrite this functionality using ``pyuvdata`` and making it more homogeneous with the rest of the functions.

### LINKED ISSUE #28: need a working notebook or script example cpu_vis
It seems that there key inputs that are required that are not generated by code inside of hera_sim (crd_eq, antenna array - which is a holdover from capo and deprecated, and should be replaced with something from pyuvdata).  A working example would be very useful.

steven-murray:
This is not quite working perfectly yet. I have attempted to convert the ``vis_cpu`` function into a class which has an interface that I *think* could be general enough for several different simulators. I've also added a notebook tutorial on how to use it.

It runs without error, but I don't necessarily trust the output. It would make testing much faster if I could compare exactly to an example usage of the old code. Hopefully @AaronParsons  can help here. 

There is definitely something strange with the beam mapping. 

Looking for input as to

1. Whether the general API is adequately general for our purposes.
2. How to get ``VisCPU`` to exactly match ``vis_cpu``. 

Fixes #28 
Fixes #26


AaronParsons:
I don't have a problem with a class being built over vis_cpu, but I have a strong preference that the vis_cpu function be preserved verbatim.  The only reason I wrote vis_cpu was to have a function against which I could test vis_gpu, which is the GPU accelerated version.  The interfaces were chosen to match each other and one was supposed to be a drop-in replacement for the other.  Obviously, being on a GPU, we can't handle all of the magical stuff that goes into the python class you made.  My recommendation is that you keep the class, but move the "simulate" back to have it call a low-level vis_cpu function.

steven-murray:
@AaronParsons that sounds perfectly reasonable. I wasn't aware that the vis_cpu was attempting to replicate a GPUified function. 

Could you perhaps quickly paste a few lines of code with a semi-realistic call to vis_cpu, so that I can check my wrapper? I need to make sure that my new AIPY-less functions to calculate the co-ordinate transforms are behaving properly.

Also, are we keen to stick with the name ``vis_cpu``? That makes sense in relation to ``vis_gpu``, but as a standalone vis simulator, it doesn't really say much!

steven-murray:
In b8b2a1612169bc9ce08aa11bd4d633e97bd8bced I updated to keep the original ``vis_cpu`` function, with the class wrapper implementing all the new functionality I had introduced.

steven-murray:
@aelanman  and @nkern could you take a look at the updates in the newest commit, which adds basic support for healvis? I would be grateful for any suggestions on how to use it better. There are definitely some features that I have left out.

hera_sim/visibilities/healvis_wrapper.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+class HealVis(VisibilitySimulator):
+    def __init__(self, fov=180, nprocesses=1, sky_ref_chan=0, **kwargs):
+        self.fov = fov
+        self._nprocs = nprocesses
+        self._sky_ref_chan = sky_ref_chan
+
+        # A bit of a hack here because healvis uses its own AnalyticBeam,
+        # and doesn't check if you are using pyuvsim's one. This should be fixed.
```
aelanman: It's probably worth discussing this too... I'm planning enable support for analytic beams in `pyuvdata.UVBeam` and have pyuvsim and healvis both use that instead (since both already carry `pyuvdata` as a dependency). The main reason we redefined analytic beams in `healvis` is because the analytic beams in pyuvsim are strictly E-field, and `healvis` only uses power beams.

hera_sim/visibilities/simulators.py
```diff
@@ -137,40 +163,59 @@ def convert_healpix_to_point_sources(hmap):
...
-            return 0
+    def write_config_file(self, **kwargs):
+        """
+        Write out a YAML config file corresponding to the current UVData
+        object.
 
+        Args:
+            **kwargs: any options are passed to :func:`uvdata_to_config_file`.
+        """
+        uvdata_to_config_file(self.uvdata, **kwargs)
```
aelanman: You may want to note that this will write out some dummy values for certain things that can't be specified by a UVData object. Specifically -- catalog info, paths to telescope configuration files (which includes antenna layouts and the beam dictionary) and the "filing" parameters (output filename/ etc.).

There is a separate function to write telescope configuration files from a given UVData object, `uvdata_to_telescope_config`.

aelanman:
@steven-murray I've left a couple comments. Looks good! I'd be careful about using the `uvdata_to_config` functions for both healvis and pyuvsim -- While the config files are very similar, healvis needs some other options that aren't part of pyuvsim just yet. These are things like `Nside` and `FoV`, which are only needed because we're simulating healpix maps. Ideally we'll have something in pyuvsim that can handle all of this and import it.

hera_sim/visibilities/healvis_wrapper.py
steven-murray: Is this enabled yet? Can I transition this code?

hera_sim/visibilities/simulators.py
steven-murray: Thanks. I've change to using ``uvdata_to_telescop_config`` in 9abbd6f53885fa7e32b89ddbaa7acbbc60de3bf6

steven-murray: by the way, I think I should point out that I find the reading/writing of config files in pyuvsim to be hard to follow. It could be a lot more modular (like how about a separate function for creating the beam file, instead of putting that logic inside another function?).

Perhaps I should try my hand at a pull request over there :-)

steven-murray:
> I'd be careful about using the `uvdata_to_config` functions for both healvis and pyuvsim -

Hmmm, I see what you mean. The output config will not be complete (i.e. you can't read it in and run HealVis again deterministically). I'm not sure what to do about it at this stage. As you say, it seems that the `uvdata_to_config` function (and its inverse) is just not quite general enough to deal with the varied cases that might arise. 

To me, it sounds like ``Nside`` should really be a part of the ``SkyModel`` definition. ``FoV`` I guess should ideally be part of the base simulator, as it is really like a "resolution" parameter.

Anyway, I've made some changes to the code, and have some better tests (of both HealVis and VisCPU). I think there should be some more tests, but at this point I'm struggling for ideas.

Also, I could use your help if you have the time, to fix a couple of the tests that are failing, especially for HealVis. Only one is failing for healvis, which is where I have a single source which rotates on the sky for a full day, and I check that about half the times have an autocorrelation equal to the magnitude of the source (i.e. 1). 

Since healvis doesn't directly support point sources, the ``VisibilitySimulator`` wrapper class uses its default method for converting from point-sources to healpix (i.e., find the pixel that each source occupies and add the flux of that source to the pixel). However, the final result seems like it's giving me zeros for all visibilities.

Curiously, the sky intensity _does_ have a single pixel with a non-zero value in it (that is, the sky intensity that's passed to healvis). I've tried changing the latititude of the telescope and the declination of the source. Since the sky makes a full rotation, you'd expect it to show up in at least _some_ visibilities.

steven-murray:
@AaronParsons sorry to drag you into this again, but I'm wondering if you have any ideas why ``vis_cpu`` would give an autocorrelation of a uniform sky that's 90% of the expected value? As in, populating a healpix map with 1's, scaling by the pixel area, then running ``vis_cpu``, should give an autocorrelation of 0.5 (since the telescope sees half the sky). However, it gives about 0.45. This is insensitive to the healpix resolution.

steven-murray:
@Jackmastr can you post those figures comparing healvis and vis_cpu again? Maybe try getting the two simulators on the same axis?

@aelanman we wouldn't mind some help getting these to line up, though I'm assuming the fault is in how vis_cpu is wrapped.

Jackmastr:
2 telescopes, point source at zenith:
image

1 telescope, point source at zenith, "airy" beam of diameter=2:
image

steven-murray:
Thanks @Jackmastr . I'm currently going through and rebasing this on master... there have been a _lot_ of changes on both sides so it's a bit involved.

steven-murray:
OK this has been rebased onto master, and most broken tests fixed (except those in `test_vis`). 

@Jackmastr in the future I would recommend _not_ making massive sweeping changes like running `futurize` on the code in the middle of a branch that is supposed to do other things. Try to keep your commits as small and tidy as possible. 

Anyway, we can now focus on getting the visibility simulators tested properly.

Jackmastr:
One test outside of `test_vis` (`test_defaults.py::test_beam_poly_changes`) wasn't fixed for Python 3 as a heads up

Jackmastr:
This is the Travis build from when you last pushed. That test did not pass then.
> https://travis-ci.com/HERA-Team/hera_sim/jobs/237354709

CHANGELOG.rst
```diff
@@ -1,3 +1,59 @@
 =========
 Changelog
 =========
+
+
+Unreleased
+==========
+
+Added
```
steven-murray: Please add an entry here describing the new visibility simulator class!

environment.yml
```diff
@@ -5,21 +5,25 @@ channels:
...
+  - numba
   - future #only for python 2
   - pip:
     - cached-property
-    - pyuvsim
+    - git+https://github.com/RadioAstronomySoftwareGroup/pyuvsim.git
+    - git+https://github.com/RadioAstronomySoftwareGroup/pyuvdata.git
     - git+https://github.com/HERA-Team/uvtools.git
+    - git+https://github.com/radioastronomysoftwaregroup/healvis 
+    - astropy-healpix
```
steven-murray: Is this required? Shouldn't it be automatically installed via `pyuvdata`?

hera_sim/visibilities/conversions.py
```diff
@@ -0,0 +1,127 @@
+"""
+Provides a number of mappings which may be useful for visibility simulators.
+"""
+import healpy
+import numpy as np
+
+
+# def beam_healpix_to_top(hmap, n_pix_lm=63, nest=False):
```
steven-murray: let's remove this commented function

hera_sim/visibilities/conversions.py
```diff
@@ -0,0 +1,127 @@
...
+    # Get the relevant indices of res
+    bm = np.zeros((len(freqs), len(l)))
+    
+    if trunc_at_horizon:
+        bm[:, n >= 0] = efieldXX[:, n >= 0]
+    else:
+        bm = efieldXX
+    
+
+#     if trunc_at_horizon:
```
steven-murray: We can remove this dead code

hera_sim/visibilities/prisim_wrapper.py
```diff
@@ -0,0 +1,334 @@
...
+
+PRISIM_PATH = prisim.__path__[0]
+
+
+class PRISim(VisibilitySimulator):
+    diffuse_ability = False
+
+    def __init__(self, aux_file_loc='prisim_roi_file', fov_radius=90.0,
+                 beam_pol='x', beam_interp='cubic', precision='single', **kwargs):
+        self.fov_radius = fov_radius
```
steven-murray: Let's raise a `NotImplementedError` in here for now

hera_sim/visibilities/rimez_wrapper.py
```diff
@@ -0,0 +1,70 @@
...
+from .simulators import VisibilitySimulator
+from astropy import constants as cnst
+import pyuvdata
+from pyuvsim.analyticbeam import AnalyticBeam
+from RIMEz import sky_models, beam_models, rime_funcs, utils, management
+
+
+class RIMEz(VisibilitySimulator):
+
+    def __init__(self,  **kwargs):
```
steven-murray: Lets add a `NotImplementedError` in here for now

hera_sim/visibilities/vis_cpu.py
```diff
@@ -0,0 +1,248 @@
...
+
+        # Complex voltages
+        v *= A_s * Isqrt
+
+        # Compute visibilities (upper triangle only)
+        for i in range(len(antpos)):
+            np.dot(v[i: i + 1].conj(), v[i:].T, out=vis[t, i: i + 1, i:])
+
+
+    # Conjugate visibilities
```
steven-murray: Let's remove dead code

steven-murray:
I think this is super-close to merging. There are many more things that can be done, but this is working and self-contained. Just a few clean-up things, and we're good to merge.

hera_sim/visibilities/healvis_wrapper.py
```diff
@@ -0,0 +1,99 @@
...
+except ImportError:
+    raise ImportError("to use the healvis wrapper, you must install healvis!")
+
+from .simulators import VisibilitySimulator
+from astropy import constants as cnst
+
+class HealVis(VisibilitySimulator):
+    point_source_ability = False
+
+    def __init__(self, fov=180, nprocesses=1, sky_ref_chan=0, **kwargs):
```
steven-murray: I think we need to add docs fo rthese three arguments, and also make it **very clear** what it expects in terms of beams, as that has been a bit of a sticking point.

steven-murray:
Some comments form @jaguirre :

- Need to make sure we can read in the official hera antenna positions
- Have a section in the notebook simulating something like a real HERA simulation.

CHANGELOG.rst
Jackmastr: Under the "Unreleased" category or v0.1.0 [2019.08.28] or a new version?

hera_sim/visibilities/conversions.py
Jackmastr: done

hera_sim/visibilities/conversions.py
Jackmastr: done

hera_sim/visibilities/vis_cpu.py
Jackmastr: done

CHANGELOG.rst
r-pascua: I would put it under v0.1.1 [yyyy.mm.dd], following the format of the previous entry, and giving an overview of what has been added/changed/removed/etc. Also remember to update the version in the code (it lives in the VERSION file).

r-pascua: err, sorry, v0.2.0? I think we're doing semantic versioning, with major.minor.patch, so this is a new minor version, since it adds new features but doesn't break the API

environment.yml
r-pascua: I think we found in #63 that we need to include `astropy-healpix` in the environment, otherwise it never actually gets installed.

hera_sim/visibilities/healvis_wrapper.py
r-pascua: Class constructors should definitely have docstrings. Please follow the numpy docstring conventions (refer to the manual [here](https://numpydoc.readthedocs.io/en/latest/format.html)), rather than the convention that's found in most of the repository. A good reference in `hera_sim` is the `defaults` module&mdash;although it doesn't *exactly* follow numpy docstring conventions, it follows the conventions rather closely and should be a good reference. Also, make sure that constructor docstrings occur immediately after the end of the constructor code, like so:

```
class Something:
    """An example on docstrings.

    """

    def __init__(self, arg1, arg2, kwarg1="something", **kwargs):
        self.args = arg1, arg2
        self.title = kwarg1
        self.mystery = something_else(**kwargs)
    """One-liner of what it does.

    Parameters
    ----------
    arg1 : <list type(s) here>
        Provide a description of what arg1 represents.
    arg2 : list or tuple of str
        Serves as an example of how to fill in type information.
    kwarg1 : str, optional
        Something about kwarg1. Explain what the default behavior is.
    kwargs
        Passed to something_else.

    """
```

I think that the standard is to have docstrings follow the content of any of the magic methods, rather than immediately after the signature, but I'm not exactly sure&mdash;there might be a PEP guideline somewhere that addresses this. Either way, please take the time to write good docstrings, because I'd like to have the codebase employ a standard way of writing documentation by the time I finish the version 1 release.

CHANGELOG.rst
steven-murray: yup, 0.2.0

environment.yml
steven-murray: ugh. OK.

hera_sim/visibilities/healvis_wrapper.py
steven-murray: The other consideration here is that this is a subclass, and you might not want to lose the docs from the superclass.

r-pascua: Good point. Turns out I was wrong about some things in the last post&mdash;here are the corrections and additional comments.

Class constructor docstrings must be placed inside the body of the constructor definition, immediately following its signature. The above example has been updated to reflect this. Other magic methods (such as `__call__`) should have their docstrings configured in the same way.

If you want to preserve the superclass docstring, but add a new docstring for the subclass (in addition to its constructor's docstring), then you can do so by adding `self.__doc__ = ...` to the constructor&mdash;do not write a docstring for the class itself in the same way you did for the superclass, otherwise the superclass will have its docstring overwritten. Note that if you use iPython's magic `?` function, then what you see will differ depending on whether you use it on the class itself or an instance of the class. A quick example of what I mean:

```
class Base:
    """Base class docstring."""
    def __init__(self):
        """Something about the base class constructor.

        Attributes
        ----------
        A
            An arbitrarily defined parameter.

        """
        self.A = 0

class Child(Base):
    """Child class docstring. This overwrites the Base docstring."""
    def __init__(self, B):
        """Child constructor docstring.

        Parameters
        ----------
        B
            Whatever you want it to be.

        """
        self.__doc__ = "A docstring for class instances."
        self.B = B
        super().__init__()
```

You'll notice in the numpydoc guide, there's a note that classes may also have an "Attributes" section. I'm a little spotty on when exactly to use this, but it seems like it's best put into place when your class has an attribute that is not just a copy of one of its constructor's parameters. For example,
```
class Test:
    def __init__(self, x, y):
        """A silly example.

        Parameters
        ----------
        x, y : float
            Something witty.

        Attributes
        ----------
        z : float
            The product of `x` and `y`.

        """
        self.x = x
        self.y = y
        self.z = x*y
```

hera_sim/visibilities/prisim_wrapper.py
Jackmastr: Raised the error at the beginning of __init__.

hera_sim/visibilities/rimez_wrapper.py
Jackmastr: Raised the error at the beginning of __init__.

hera_sim/visibilities/healvis_wrapper.py
Jackmastr: Below is how I would write the docstring for the beginning of `VisibilitySimulator`. Would this be correct? Also I am unsure if `point_source_ability` and `diffuse_ability` should be defined in `__init__` as attributes.

```python
class VisibilitySimulator(object):
    """
    Base VisibilitySimulator class.

    Any actual visibility simulator should be sub-classed from this one.
    This class provides several convenience methods and defines the API.
    """
    
    # Whether this particular simulator has the ability to simulate point sources
    # directly
    point_source_ability = True

    # Whether this particular simulator has the ability to simulate diffuse maps
    # directly
    diffuse_ability = True

    def __init__(self, obsparams=None, uvdata=None, sky_freqs=None, beams=None,
                 beam_ids=None,
                 sky_intensity=None, point_source_pos=None, point_source_flux=None,
                 nside=2**5):
        """
        Parameters
        ----------
        obsparams : dict or filepath
            Exactly the expected input to `pyuvsim`'s
            :func:`pyuvsim.simsetup.initialize_uvdata_from_params` function.
            If provided, `uvdata`, `beams`, and `beam_ids` are not required.
        uvdata : UVData object
            A :class:`pyuvdata.UVData` object contain information about
            the "observation".
        sky_freqs : array_like
            Frequencies at which the sky intensity and/or point sources
            are defined in [Hz]. Shape=(NFREQS,).
        beams : array_like, optional
            UVBeam models for as many antennae as have unique beams.
            By default, a single uniform beam is applied for every antenna.
            Shape=(N_BEAMS,).
        beam_ids : array_like of int, optional
            List of integers specifying which beam model each antenna uses
            (i.e. the index of `beams` which it should refer to). By default,
            all antennas use the same beam (beam 0). Shape=(N_ANTS,).
        sky_intensity : array_like
            A healpix model for the intensity of the sky emission, in [Jy/sr].
            Shape=(NFREQS, N_PIX_SKY).
        point_source_pos : array_like, optional
            An array of point sources. For each source, the entries are
            (ra, dec) [rad] (assumed to be in J2000). Shape=(N_SOURCES, 2).
        point_source_flux : array_like, optional
            An array of fluxes of the given point sources, per frequency.
            Fluxes in [Jy]. Shape=(NFREQS, N_SOURCES).
        nside : int, optional
            Only used if sky_intensity is *not* given but the simulator
            is incapable of directly dealing with point sources. In this
            case, it sets the resolution of the healpix map to which the
            sources will be allocated.
        """
        if obsparams:
            self.uvdata, self.beams, self.beam_ids = initialize_uvdata_from_params(obsparams)
        else:
            if uvdata is None:
                raise ValueError("if obsparams is not given, uvdata must be")

            self.uvdata = uvdata
            self.beams = [ab.AnalyticBeam("uniform")] if beams is None else beams
            self.beam_ids = np.zeros(self.n_ant, dtype=np.int) if beam_ids is None else beam_ids

        self._nside = nside
        self.sky_intensity = sky_intensity

        self.sky_freqs = sky_freqs
        if sky_freqs is None:
            self.sky_freqs = np.unique(uvdata.freq_array)

        self.point_source_pos = point_source_pos
        self.point_source_flux = point_source_flux

        self.validate()
```

r-pascua: This looks very good, but I have a few comments/questions. 

First, I think that it would be better to include `point_source_ability` and `diffuse_ability` as attributes of the class, inside `__init__`. I'm assuming that subclasses would update these accordingly at instantiation.

Second, comments should typically not exceed 72 characters (including indentation) in order to conform to PEP 8. There are certainly times when this rule can be broken, though&mdash;see [here](https://www.python.org/dev/peps/pep-0008/) for details.

Finally, I have a question about the `sky_intensity` parameter and a comment about optional parameters. Regarding `sky_intensity`, it's not listed as optional in the docstring, similar to `obsparams` and `uvdata`, but it is not treated in the same way as `obsparams` and `uvdata`&mdash;these are not listed as optional, and an exception is raised if neither of the two are defined, somewhat justifying the omission of the "optional" label. So, is `sky_intensity` optional or required? Now, regarding optional parameters, these should have brief descriptions of their default behavior, preferably as a closing note in their description under the "parameters" section. (I don't exactly recall if optional parameters *must* have their default behavior described according to numpy docstring conventions, but I think it's extremely helpful to have it written down somewhere, for the benefit of both future developers and end-users.)

On a somewhat different note, I would like to start a conversation with @steven-murray regarding how we'll handle creation and initialization of classes in the future, given the move to Py3. Py3 allows for direct control of both the `__new__` and `__init__` magic methods, and I'm curious if we should look into exploiting these, as they serve somewhat different purposes. This definitely isn't the venue for that conversation, so I'd appreciate a suggestion about where we should have that discussion (slack? open an issue?).

steven-murray: The `point_source_ability` and `diffuse_ability` are class attributes, not instance attributes, i.e. they should remain outside the `__init__`. They are properties of the _algorithm_, not of any particular instance of it. 

`sky_intensity` is not required -- _either_ `sky_intensity` _or_ `point_source_pos` and `point_source_flux` are required (getting both is OK too).

@r-pascua is right about addressing default behaviour. If the parameter just has an obvious default set in the constructor (i.e. `def foo(bar=3): ...`, then the default behaviour can be omitted from the docstring, because Sphinx will report it properly. If it's `bar=None` and then you do something weird to it in the `__init__`, you need to specify that in the docstring.

@r-pascua we can talk about this more on Slack, but my understanding is that `__new__` is really only useful for changing the _class_ properties at instantiation, not instance properties. This has a few interesting corner-case uses (especially for plugins etc.). Not really useful here I wouldn't think.

Jackmastr: `uvdata` and `obsparams` would also be optional too, right? Since you can choose to pass one or the other, correct?

steven-murray: Yup!

steven-murray: Also, remember to make docstrings for the _subclasses_, not just the main class.

Jackmastr: Documentation has been added that should be (mostly) PEP257 compliant at this point for `simulators.py`, `healvis_wrapper.py`, `vis_cpu.py` and `conversions.py`.

I was not able to have the subclasses inherit part of  superclass docstring in the way suggested ... but I am not sure that is common practice. Users should always be able to just look at the superclass documentation for that information.

steven-murray: I think that's fine... it gets a bit finnicky to inherent docs anyway. As long as it is clear. I'll have a check now.

hera_sim/visibilities/conversions.py
```diff
@@ -90,10 +51,22 @@ def uvbeam_to_lm(uvbeam, freqs, n_pix_lm=63, trunc_at_horizon=False, **kwargs):
...
-    Ripped straight from aipy.
+    Calculates the equatorial to topocentric conversion matrix.
+    
+    Conversion at a given hour angle (ha) and declination (dec). Ripped 
+    straight from aipy.
+    
+    Parameters
+    ----------
+    ha : float
+        Hour angle.
```
steven-murray: Give units!

Jackmastr: added that they should both be radians

CHANGELOG.rst
Jackmastr: Done. Take a look.

hera_sim/visibilities/__init__.py
```diff
@@ -0,0 +1,34 @@
...
+import warnings
+
+# Registered Simulators
+from .vis_cpu import VisCPU
+
+
+try:
+    from .healvis_wrapper import HealVis
+except (ImportError, NameError) as e: # NameError arises bc healvis is not Python 3 compliant
+    warnings.warn("healvis failed to import.")
```
r-pascua: Maybe a little nitpicky, but it would be good to include similar warnings if the other simulators fail to import. If you have the `warnings.warn(msg)` statement, then there's no need to `pass` as well. I also think that you can just do 
```
try:
    something
except [Exceptions,]:
    warnings.warn(msg)
```
rather than giving the exceptions an alias (without the `as e` part); the only time you really need to do that (as far as I know) is when you intend to do something with the exception.

hera_sim/visibilities/conversions.py
```diff
@@ -0,0 +1,106 @@
...
+    -------
+    ndarray
+        Coordinate transform matrix converting equatorial coordinates to
+        topocentric coordinates. Shape=(3, 3).
+    """
+    sin_H, cos_H = np.sin(ha), np.cos(ha)
+    sin_d, cos_d = np.sin(dec), np.cos(dec)
+    zero = np.zeros_like(ha)
+
+    map = np.array([[sin_H, cos_H, zero],
```
r-pascua: I tend to shy away from using the names of builtins for variable names. I know that this only changes the reference of `map` in the scope of `eq2top_m`, but could we maybe rename this to `conversion` or something else semantically accurate that isn't overloading a builtin?

hera_sim/visibilities/healvis_wrapper.py
```diff
@@ -0,0 +1,141 @@
...
+            Frequency reference channel. Defaults to 0.
+        **kwargs
+            Arguments of :class:`VisibilitySimulator`.
+        """
+        self.fov = fov
+        self._nprocs = nprocesses
+        self._sky_ref_chan = sky_ref_chan
+
+        # A bit of a hack here because healvis uses its own AnalyticBeam, and
+        # doesn't check if you are using pyuvsim's one. This should be fixed.
```
r-pascua: Can you make an issue noting that this should be fixed?

hera_sim/visibilities/healvis_wrapper.py
```diff
@@ -0,0 +1,141 @@
...
+            spectral_index = old_args['spectral_index']
+            diameter = old_args['diameter']
+            kwargs['beams'] = [AnalyticBeam(beam_type=beam_type,
+                                            gauss_width=gauss_width,
+                                            diameter=diameter,
+                                            spectral_index=spectral_index)]
+
+        super(HealVis, self).__init__(**kwargs)
+
+    def validate(self):
```
r-pascua: Maybe leave a note here for future developers that this just augments the behavior of the superclass `validate` method for instances of `HealVis`? I was thrown off at first when I saw that a `validate` method was defined but never called in this class.

hera_sim/visibilities/simulators.py
```diff
@@ -0,0 +1,333 @@
...
+        return np.array([ra*np.pi/180, dec*np.pi/180]).T, flux
+
+    def simulate(self):
+        """Perform the visibility simulation."""
+        self._write_history()
+        vis = self._simulate()
+        self.uvdata.data_array += vis
+        return vis
+
+    def _simulate(self):
```
r-pascua: Make this an `abstractmethod`. In case you don't know the proper way to do so, here's a quick rundown:

1. add to your imports: `from abc import ABCMeta, abstractmethod`
2. update the class declaration to `class VisibilitySimulator(object, metaclass=ABCMeta):`
3. decorate the `_simulate` method with `@abstractmethod`

Voila! This will enforce that every subclass of `VisibilitySimulator` *must* define a `_simulate` method.

hera_sim/visibilities/vis_cpu.py
```diff
@@ -0,0 +1,328 @@
...
+
+        Notes
+        -----
+            Due to using the verbatim :func:`vis_cpu` function, the beam
+            cube must have an entry for each antenna, which is a bit of
+            a waste of memory in some cases. If this is changed in the
+            future, this method can be modified to only return one
+            matrix for each beam.
+        """
+        return np.array([
```
r-pascua: If you're concerned about memory, then you might want to use `np.asarray` instead. `np.array` defaults to make a copy of the object to be made into an `ndarray` (you could also just do `copy=False`, but I think using `np.asarray` is cleaner).

hera_sim/visibilities/vis_cpu.py
```diff
@@ -0,0 +1,328 @@
...
+        Calculate approximate HEALPix map of point sources.
+
+        Returns
+        -------
+        array_like
+            equatorial coordinates of Healpix pixels, in Cartesian
+            system. Shape=(3, NPIX).
+        """
+        ra, dec = self.point_source_pos.T
+        return np.array([np.cos(ra)*np.cos(dec), np.cos(dec)*np.sin(ra),
```
r-pascua: Again, might want to consider using `np.asarray` here instead.

CHANGELOG.rst
```diff
@@ -3,12 +3,29 @@ Changelog
...
 
-v0.2.0 [2019.11.20]
+v0.3.0 [2019.12.10]
 ===================
 
 Added
 -----
+- New sub-package ``simulators``
+    - ``VisibilitySimulators`` class
+        - Provides a common interface to interferometric visibility simulators.
```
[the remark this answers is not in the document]
steven-murray: Seems like it looks OK: https://github.com/HERA-Team/hera_sim/blob/visibility-simulators/CHANGELOG.rst

docs/tutorials/example_antpos_uvsim.csv
```diff
@@ -0,0 +1,6 @@
+Name        Number   BeamID   E        N       U
```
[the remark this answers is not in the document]
steven-murray: Put them in `data/` in the `tutorial` folder.

docs/tutorials/obsparam_example.yml
```diff
@@ -0,0 +1,18 @@
+freq:
```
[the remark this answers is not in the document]
steven-murray: Changed to YAML.

hera_sim/io.py
```diff
@@ -47,7 +47,7 @@ def empty_uvdata(nfreq, ntimes, ants, **kwargs):
         A :class:`pyuvdata.UVData` object, unfilled.
 
     """
-    start_time = kwargs.pop("start_time", 2458119.5)
+    start_time = kwargs.pop("start_time", 2456658.5) # January 1 2014
```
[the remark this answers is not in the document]
steven-murray: Otherwise we get IERS errors... massive rabbit hole.

hera_sim/tests/test_noise.py
```diff
@@ -33,15 +33,17 @@ def test_resample_Tsky(self):
         # import uvtools, pylab as plt
         # uvtools.plot.waterfall(tsky); plt.show()
 
+### Does not pass, needs to be commented for the pull request
```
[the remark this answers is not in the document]
steven-murray: I remember making the times greater at some point (probably after this branch was branched). I'll fix that and see if it works.

docs/tutorials/example_telescope_uvsim.yaml
```diff
@@ -0,0 +1,6 @@
+beam_paths:
```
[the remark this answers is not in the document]
steven-murray: Done

hera_sim/tests/test_noise.py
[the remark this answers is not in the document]
steven-murray: I have reverted to the test code as it is on master and it works fine.

hera_sim/visibilities/__init__.py
steven-murray: I've given them all a warning, and removed redundant `e`.

hera_sim/visibilities/healvis_wrapper.py
steven-murray: I think this really needs to be fixed on `healvis` rather than here. I think @aelanman is already on it.

hera_sim/visibilities/healvis_wrapper.py
steven-murray: I've made the docstring a bit more descriptive. Does that help?

hera_sim/visibilities/simulators.py
steven-murray: Done (except with py2 compatibility for now).

hera_sim/visibilities/vis_cpu.py
steven-murray: Cool, thanks, didn't know that!

hera_sim/visibilities/vis_cpu.py
steven-murray: Done.

PR #39, opened 2019-04-12
dtv_rfi

steven-murray:
Digital TV RFI (suggested from https://github.com/HERA-Team/hera_qm/issues/270). Currently implements single-LST, definite-bandwidth random RFI in a given overall band. Defaults set to what @mwilensky768 suggested in the issue referenced above. 

I may have misunderstood some of the details that @mwilensky768 gave, so would appreciate a good look-over.

### DIFF
--- a/hera_sim/rfi.py
+++ b/hera_sim/rfi.py
@@ -238,11 +238,16 @@ def rfi_dtv(fqs, lsts, rfi=None, freq_min=.174, freq_max=.214, width=0.008,
 
     bands = np.arange(freq_min, freq_max, width)  # lower freq of each potential DTV band
 
+    # If the bands fit exactly into freqs, the upper band will be the top freq
+    # and we need to ignore it.
+    if fqs.max() <= bands.max():
+        bands = bands[:-1]
+
     delta_f = fqs[1] - fqs[0]
 
-    chance = list(chance)
-    strength_std = list(strength_std)
-    strength = list(strength)
+    chance = _listify(chance)
+    strength_std = _listify(strength_std)
+    strength = _listify(strength)
 
     if len(chance) == 1:
         chance *= len(bands)
@@ -259,13 +264,30 @@ def rfi_dtv(fqs, lsts, rfi=None, freq_min=.174, freq_max=.214, width=0.008,
         raise ValueError("strength_std must be float or list with len equal to number of bands")
 
     for band, chnc, strngth, str_std in zip(bands, chance, strength, strength_std):
-        fq_ind_min = np.argwhere(band >= fqs)[0][0]
-        fq_ind_max = fq_ind_min + int(width / delta_f)
-        this_rfi = rfi[:, fq_ind_min:fq_ind_max]
+        fq_ind_min = np.argwhere(band <= fqs)[0][0]
+        fq_ind_max = fq_ind_min + int(width / delta_f) + 1
+        this_rfi = rfi[:, fq_ind_min:min(fq_ind_max, fqs.size)]
 
         rfis = np.random.uniform(size=lsts.size) <= chnc
-        this_rfi.flat[rfis] += np.random.normal(strngth, str_std, size=np.sum(rfis)) * np.exp(
+        this_rfi[rfis] += np.atleast_2d(np.random.normal(strngth, str_std, size=np.sum(rfis)) * np.exp(
             2 * np.pi * 1j * np.random.uniform(size=np.sum(rfis))
-        )
+        )).T
 
     return rfi
+
+
+def _listify(x):
+    """
+    Ensure a scalar/list is returned as a list.
+
+    Gotten from https://stackoverflow.com/a/1416677/1467820
+    """
+    if isinstance(x, basestring):
+        return [x]
+    else:
+        try:
+            iter(x)
+        except TypeError:
+            return [x]
+        else:
+            return list(x)
--- a/hera_sim/rfi.py
+++ b/hera_sim/rfi.py
@@ -228,7 +228,7 @@ def rfi_dtv(fqs, lsts, rfi=None, freq_min=.174, freq_max=.214, width=0.008,
     """
     if rfi is None:
         rfi = np.zeros((lsts.size, fqs.size), dtype=np.complex)
-    assert rfi.shape == (lsts.size, fqs.size)
+    assert rfi.shape == (lsts.size, fqs.size), "rfi shape is not (lst.size, fqs.size)"
 
     bands = np.arange(freq_min, freq_max, width) # lower freq of each potential DTV band
 
--- a/hera_sim/rfi.py
+++ b/hera_sim/rfi.py
@@ -240,7 +240,7 @@ def rfi_dtv(fqs, lsts, rfi=None, freq_min=.174, freq_max=.214, width=0.008,
 
     delta_f = fqs[1] - fqs[0]
 
-    chance = list(change)
+    chance = list(chance)
     strength_std = list(strength_std)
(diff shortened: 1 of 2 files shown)


hera_sim/rfi.py
```diff
@@ -190,3 +190,58 @@ def rfi_scatter(fqs, lsts, rfi=None, chance=0.0001, strength=10, std=10):
...
+    delta_f = fqs[1] - fqs[0]
+
+    for band in bands:
+        fq_ind_min = np.argwhere(band >= fqs)[0][0]
+        fq_ind_max = fq_ind_min + int(width/delta_f)
+        this_rfi = rfi[:, fq_ind_min:fq_ind_max]
+
+        rfis = np.where(np.random.uniform(size=this_rfi) <= chance)[0]
+        this_rfi.flat[rfis] += np.random.normal(strength, strength_std, size=rfis.size)*np.exp(
+            1j*np.random.uniform(size=rfis.size)
```
mwilensky768: Do you mean to confine the phase of the signal to be between 0 and 1 (or pi and pi+1 for when the initial gaussian draw is negative)? To get all phases between 0 and 2pi, this line can go two different ways.

```suggestion
            1j*np.random.uniform(0, 2*np.pi, size=rfis.size)
```

or

```suggestion
            2 * np.pi * 1j*np.random.uniform(size=rfis.size)
```

This also happens in the code for rfi_scatter() (line 190 on master I think - not sure how to link from here)

steven-murray:
thanks @mwilensky768 , good spotting. Scary that those errors have existed in other RFI functions for who knows how long...

hera_sim/rfi.py
```diff
@@ -184,9 +189,65 @@ def rfi_scatter(fqs, lsts, rfi=None, chance=0.0001, strength=10, std=10):
...
+        rfi (array-like): shape=(NTIMES,NFREQS), default=None
+            an array to which the RFI will be added.  If None, a new array
+            is generated.
+        freq_min, freq_max (float):
+            the min and max frequencies of the full DTV band [GHz]
+        width (float):
+            Width of individual DTV bands [GHz]
+        chance (float): default=0.0001
+            the probability that a time/freq bin will be assigned an RFI impulse
+        strength (float): Jy, default=10
```
adampbeardsley: Is this default (and strength_std below) in line with what @mwilensky768 sees in data?
In general we should evaluate some of these options at some point.

adampbeardsley: Also, should we allow for an array input for the different channels?

hera_sim/rfi.py
```diff
@@ -184,9 +189,65 @@ def rfi_scatter(fqs, lsts, rfi=None, chance=0.0001, strength=10, std=10):
...
+    bands = np.arange(freq_min, freq_max, width)  # lower freq of each potential DTV band
+
+    delta_f = fqs[1] - fqs[0]
+
+    for band in bands:
+        fq_ind_min = np.argwhere(band >= fqs)[0][0]
+        fq_ind_max = fq_ind_min + int(width / delta_f)
+        this_rfi = rfi[:, fq_ind_min:fq_ind_max]
+
+        rfis = np.where(np.random.uniform(size=this_rfi) <= chance)[0]
```
adampbeardsley: I think this results in each channel within a band acting independently. But I think we should have a given band either all on or all off at the same time.

hera_sim/rfi.py
mwilensky768: I actually thought about the strength/strength_std myself, but declined to comment in my review. I'll explain below.

As set here, the defaults are saying that the RFI-signal/RFI-noise ratio is 1. I certainly don't see that in SSINS, but the SSINS are a highly averaged-down space. Since I don't really work with the raw visibilities, I don't actually know what the RFI-signal/RFI-noise ratio is in the raw visibilities. Seeing as it gets caught by xrfi (single-baseline flagger) some of the time, I would hazard to guess that the ratio of RFI-signal/RFI-noise is greater than 1 in the raw visibilities.

bhazelton: At least in the MWA, it's not uncommon for it to be undetectable in single visibilities (right @mwilensky768?)

hera_sim/rfi.py
mwilensky768: This is a very good catch @adampbeardsley. I _think_ if you set up a boolean index on the lsts like so:

```suggestion
        rfis = np.random.uniform(size=lsts.size) <= chance
```

and then replace line 249 with

```
# shape of subarray that is called by boolean index
sz = [np.count_nonzero(rfis), int(width / delta_f)]
this_rfi[rfis] += np.random.normal(strength, strength_std, size=sz) * np.exp(
	            2 * np.pi * 1j * np.random.uniform(size=sz)
```

that ought to work. Please let me know your thoughts.

mwilensky768: I'm worried about the second component of the sz vector. Might need to add 1 to it.

hera_sim/rfi.py
adampbeardsley: Hmm, couple thoughts here:
1) I'm not sure this is really a signal to noise ratio. It's a signal vs variation of the signal -- so sometimes it's bright and other times it's not. This is different than say thermal noise of the visibilities.
2) The xrfi we're running now (and where I've looked into this a little) is averaged across baselines (in a weird way), so not a single-baseline flagger. I also haven't looked back to the raw visibilities though.

mwilensky768: I guess that since sometimes it is detectable in the single visibilities and sometimes not, that a SNR of 1 is a good middle-ground default.

hera_sim/rfi.py
```diff
@@ -1,7 +1,7 @@
 """A module for generating realistic HERA RFI."""
 
-import numpy as np
 import aipy
```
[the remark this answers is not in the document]
steven-murray: it is, actually, but I've swapped it for astropy in 4fd2c2ed97bdc8e4cbcbb03956250bd50c4d17cd

hera_sim/rfi.py
steven-murray: Defaults were copied from ``rfi_scatter``, as I have no feel for what they should be. I think @adampbeardsley  is right that this is a good project.

I've updated code to allow array input for different bands in 62e00855410610f1326b5394411926f5a49e205a.

mwilensky768: @adampbeardsley if the different frequencies are drawn independently, then I'm not seeing the difference between variation on a signal and noise on a signal. Happy to table that for now though.

hera_sim/rfi.py
steven-murray: I've addressed this in 342b0a697db2f8afa857058af6da6bf3dbbd8c7a

steven-murray:
I've added some tests in the latest commit -- found a few bugs in doing so. Now it seems to have roughly the expected behaviour (contiguous frequencies receive the same strength within a band, different bands are independent, each LST is independent). 

@mwilensky768 the phase rotation is interesting. Nothing is doing that at the moment, and it could be added if people think it is a good idea?

hera_sim/rfi.py
steven-murray: I think what @adampbeardsley is getting at is that there is thermal noise in the visibility (potentially), and the std in this routine does not specify _that_ level of noise, just how much variation there is between levels of RFI "signal" for different LSTS/FREQS

adampbeardsley:
I think phase rotation would require a DFT to baseline location. Probably useful for realistic sims, but maybe not when we're using it to test RFI flagging, which doesn't use the phase info.

PR #40, opened 2019-04-12
Installation issues

### LINKED ISSUE #38: add cached-property as a dependency
This does not seem to be standard in anaconda installs.

`(hera) -bash-4.2$ python setup.py install
Traceback (most recent call last):
  File "setup.py", line 5, in <module>
    from hera_sim import version
  File "/lustre/aoc/projects/hera/jaguirre/gitrepos/hera_sim/hera_sim/__init__.py", line 11, in <module>
    from . import simulate
  File "/lustre/aoc/projects/hera/jaguirre/gitrepos/hera_sim/hera_sim/simulate.py", line 12, in <module>
    from cached_property import cached_property
ImportError: No module named cached_property
(hera) -bash-4.2$ conda install cached-property`

steven-murray:
Fixes #38 

Fixes a few issues related to installation. Also adds more documentation in README for installation. It may be best long-term to create a Conda package, but I've never done that.

I thought @aelanman  might want to have a look at how I solved the requirements problem in setup.py. May or may not be applicable to the healvis installation issue: https://github.com/RadioAstronomySoftwareGroup/healvis/issues/19

### DIFF
--- a/hera_sim/foregrounds.py
+++ b/hera_sim/foregrounds.py
@@ -6,7 +6,6 @@ at the requested lsts and frequencies.
 """
 
 import numpy as np
-from scipy.interpolate import RectBivariateSpline
 from aipy.const import sidereal_day
 
 from . import noise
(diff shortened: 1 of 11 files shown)


hera_sim/foregrounds.py
```diff
@@ -118,7 +117,7 @@ def pntsrc_foreground(lsts, fqs, bl_vec, nsrcs=1000, Smin=0.3, Smax=300,
         dtau = np.random.uniform(-.1 * bl_len_ns, .1 * bl_len_ns)  # XXX adds a bit to total delay, increasing bl_len_ns
         vis[t, :] += flux * (fqs / mfreq) ** index * np.exp(2j * np.pi * fqs * dtau)
     ha = utils.compute_ha(lsts, 0)
-    for fi in xrange(fqs.size):
+    for fi in range(fqs.size):
```
mkolopanis: I'd recommend using a forwards compatibility on range in python 2. something like: 
`from builtins import range`
which brings the python 3 `range` feature into python 2. In py2 `range` returns a list which can cause issues if your iterator is very large. `six.moves` also has a range import too, but I'm not sure what is listed as a dependency for this repo.

steven-murray: Fixed this in a67c4f71c5b8bb38be98f811fb70e890acb2402f

PR #43, opened 2019-06-09
frequency evolution in reflection parameters

nkern:
Enables specifying frequency evolution in reflection parameters


nkern:
@steven-murray yup I agree that is one particular silent error mode, and I wasn't exactly sure how to handle it best. I think its fine as is, but if you prefer a warning to be issued we can do that!

steven-murray:
Let's add the warning -- if one knows that it's going to happen, and don't want the warning displayed, they can always filter it. At least this way it's harder to go wrong.

steven-murray:
Thanks @nkern this looks good.

PR #45, opened 2019-06-18
Move logic for empty_uvdata to pyuvsim.simsetup

steven-murray:
This PR seeks to simplify the `io.empty_uvdata` function so that it basically just uses `pyuvsim`. 

I have checked that all tests work (some tests required removal, since there is now no function for calculating antpairs, and others required fixing). 

I think more can be done to simplify this. It seems to me that `pyuvsim` (or even `pyuvdata`!) should have its own function that does the job of `init_uvdata_out` without all the extra filenames etc., and that this should be callable from within the pyuvsim function. This would remove an extra function here.

Furthermore, it seems that the `initialize_uvdata_from_keywords` sets up the uvdata object with a tuple for the `telescope_location`, which fails the `check()`. Converting to a list fixes it. This is done in this PR, but should be in `pyuvsim`.

Also along these lines, it seems that the `antenna_names` aren't passed through properly, and they have to be set manually. 

Finally, the `polarization_array` is hard-coded to be `[-5, -6, -7, -8]` in `pyuvsim`, which I don't think is a good idea (but I might be wrong here).

### DIFF
--- a/hera_sim/tests/test_noise.py
+++ b/hera_sim/tests/test_noise.py
@@ -35,13 +35,13 @@ class TestNoise(unittest.TestCase):
 
     def test_sky_noise_jy(self):
         fqs = np.linspace(0.1, 0.2, 100)
-        lsts = np.linspace(0, 2 * np.pi, 300)
+        lsts = np.linspace(0, 2 * np.pi, 500)
         omp = noise.bm_poly_to_omega_p(fqs)
         tsky = noise.resample_Tsky(fqs, lsts)
         jy2T = noise.jy2T(fqs, omega_p=omp) / 1e3
         jy2T.shape = (1, -1)
         nos_jy = noise.sky_noise_jy(tsky, fqs, lsts, inttime=10.7, omega_p=omp)
-        self.assertEqual(nos_jy.shape, (300, 100))
+        self.assertEqual(nos_jy.shape, (500, 100))
         np.testing.assert_allclose(np.average(nos_jy, axis=0), 0, atol=0.7)
         scaling = np.average(tsky, axis=0) / jy2T
         np.testing.assert_allclose(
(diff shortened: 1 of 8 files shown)


steven-murray:
OK. This will need to wait for `pyuvsim` to be updated on `PyPI` before it will work.

hera_sim/tests/test_noise.py
```diff
@@ -50,7 +50,7 @@ def test_sky_noise_jy(self):
         np.random.seed(0)
         nos_jy = noise.sky_noise_jy(tsky, fqs, lsts, inttime=None, omega_p=omp)
         np.testing.assert_allclose(
-            np.std(nos_jy, axis=0) / scaling * np.sqrt(1e6 * aipy.const.sidereal_day/200), 1.0, atol=0.1
+            np.std(nos_jy, axis=0) / scaling * np.sqrt(1e6 * aipy.const.sidereal_day/300), 1.0, atol=0.1
```
nkern: this could be the issue

nkern:
@steven-murray I agree with the things you point out. I think easy, near-trivial fixes in `pyuvdata` are converting the telescope location to a list and making `polarization_array` a kwarg. Could you make an issue on pyuvdata for this? I think making an `init_uvdata` function in `pyuvdata` is the way to go, but probably requires a little more work. I'm pretty happy w/ what you have here, I'll try to figure out what's wrong with the tests in the meantime..

hera_sim/tests/test_noise.py
nkern: yup this was it

steven-murray:
@nkern actually I think the relevant changes have been made in `pyuvsim` and things seem to work now. I agree, having an `init_uvdata` function in `pyuvdata` seems to make sense. I'll make an issue over there. Thanks for your review!

PR #46, opened 2019-07-18
Make data

r-pascua:
I added a `make_data` function as a member function of the `Simulator` class. The gist of how it works is as follows: the user provides either a dictionary or hook to a YAML file as an argument, and all the desired simulation components (specified in the `sim_params` argument) are added to the `Simulator` object that called the function. I have not added any tests for the `make_data` function, as I'm a little uncertain on what sort of simple tests we can run on it.


hera_sim/simulate.py
```diff
@@ -434,4 +435,64 @@ def add_xtalk(self, model='gen_whitenoise_xtalk', bls=None, **kwargs):
...
+            
+            sim_params (dict): dictionary of simulation parameters.
+                it must take the form {model:params}, where model is
+                a string specifying the simulation component to be added
+                and params is a dictionary providing all the keyword
+                arguments and their desired values.
+        """
+        # make a dictionary whose values point to the various methods
+        # used to add different simulation components
+        SIMULATION_COMPONENTS = { 'noiselike_eor':self.add_eor,
```
nkern: lets put this at the top of the module in the global namespace

hera_sim/simulate.py
```diff
@@ -434,4 +435,64 @@ def add_xtalk(self, model='gen_whitenoise_xtalk', bls=None, **kwargs):
...
+                'simulation parameters must be provided.'
+
+        # if a hook to a simulation file is provided, then read it in
+        # XXX is there a way to check that the formatting of sim_file is OK?
+        if sim_file is not None:
+            with open(sim_file, 'r') as doc:
+                sim_params = yaml.load(doc.read(), Loader=yaml.FullLoader)
+        
+        # enforce that sim_params are of the correct form
+        assert type(sim_params) is dict, \
```
nkern: lets make this `assert isinstance(sim_params, dict)`

hera_sim/simulate.py
```diff
@@ -434,4 +435,64 @@ def add_xtalk(self, model='gen_whitenoise_xtalk', bls=None, **kwargs):
...
+                'Simulation parameters must be passed as a dictionary.'
+
+        for model, params in sim_params.items():
+            assert model in SIMULATION_COMPONENTS.keys(), \
+                    'Models must be supported by hera_sim. ' + \
+                    "'{}' is currently not supported.".format(model)
+
+            assert type(params) is dict, \
```
nkern: same here

hera_sim/simulate.py
```diff
@@ -434,4 +435,64 @@ def add_xtalk(self, model='gen_whitenoise_xtalk', bls=None, **kwargs):
                 vis=self.data.data_array[blt_ind, 0, :, pol_ind],
                 xtalk=xtalk
             )
+    
+    def make_data(self, sim_file=None, sim_params=None):
```
nkern: if you want you can make this just `make_data(self, sim_params)`
and do
```
if isinstance(sim_params (str, np.str)):
    with open(sim_params, 'r') as doc:
        sim_params = yaml.load(doc.read(), Loader=yaml.FullLoader)
```

hera_sim/simulate.py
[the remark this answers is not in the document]
r-pascua: Do you know if there's a sort of plugin that can search through all the other `hera_sim` modules and find all of the possible simulation components? I'm envisioning having something that goes and looks through, say `eor`, `foregrounds`, etc. and pulls out the different simulation functions, then does this sort of mapping where the name maps to the `Simulator` method.

hera_sim/simulate.py
```diff
@@ -434,4 +435,64 @@ def add_xtalk(self, model='gen_whitenoise_xtalk', bls=None, **kwargs):
...
+        all of the desired simulation components to the Simulator object.
+
+        Args:
+            sim_file (str): string providing a hook to a YAML file
+            
+            sim_params (dict): dictionary of simulation parameters.
+                it must take the form {model:params}, where model is
+                a string specifying the simulation component to be added
+                and params is a dictionary providing all the keyword
+                arguments and their desired values.
```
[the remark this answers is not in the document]
r-pascua: I'm not quite sure what you mean by your second comment, but as for the first one, `sim_file` will always override `sim_params` if both are provided in the current implementation. See lines 473-475.

hera_sim/simulate.py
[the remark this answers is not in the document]
r-pascua: I agree that it's better to generate this mapping automatically, but I currently have no idea how to go about implementing that. I'll read up on plugins to see if I can find something that will do the trick, and I'll implement it if I find something that works.

hera_sim/simulate.py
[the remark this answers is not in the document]
r-pascua: To be honest, I'm not exactly sure what i was getting at with the question of checking the formatting of the YAML file. I was mainly concerned about the safety of `yaml`'s `FullLoader`, although if I remember correctly, the developers state that the `FullLoader` prevents execution of code.

hera_sim/simulate.py
```diff
@@ -434,4 +435,64 @@ def add_xtalk(self, model='gen_whitenoise_xtalk', bls=None, **kwargs):
...
+
+        # make a tuple keeping track of which components do not use a model
+        uses_no_model = ('gains', 'sigchain_reflections')
+
+        assert sim_file is not None or sim_params is not None, \
+                'Either a hook to a simulation file or a dictionary of ' + \
+                'simulation parameters must be provided.'
+
+        # if a hook to a simulation file is provided, then read it in
+        # XXX is there a way to check that the formatting of sim_file is OK?
```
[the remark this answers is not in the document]
r-pascua: I admit that I'm not exactly sure what you mean by this, since I was referencing `sim_file`. That said, I'm not exactly sure what i was getting at with the question of checking the formatting of the YAML file. I was mainly concerned about the safety of `yaml`'s `FullLoader`, although if I remember correctly, the developers state that the `FullLoader` prevents execution of code.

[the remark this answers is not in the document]
r-pascua: Is it that you were suggesting I `try` to load in the contents of `sim_file` and do something else if the `except` clause is executed?

r-pascua:
I implemented most of the changes recommended. I kept the function arguments the same, but I think that was about the only thing I didn't change that was suggested. Give it another look and let me know what you think.

hera_sim/simulate.py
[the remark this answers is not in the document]
steven-murray: Hmmm, I think if anything it should be the other way round. That is, either error if both are passed, or let the `sim_params` override the `sim_file` (and document that). `sim_params` is more dynamic than the file, and so the user is more likely to override the file with params than the other way around.

hera_sim/simulate.py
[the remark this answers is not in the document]
steven-murray: So there's no way we can do this well in the current implementation. You can definitely go through each module and find which functions are defined in it -- but there's no automatic way to tell if those functions correspond to "models", nor if they correspond to a prescribed API. 

I think the "best" way to achieve this is to define each model as a class, which has a metaclass that keeps track of its subclasses (i.e. the "plugins" that adopt that API). This would take a bit of setup outside the scope of this PR though. Let's try get this one accepted, and open an issue for it.

hera_sim/simulate.py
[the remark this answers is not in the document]
steven-murray: yeah, that's what I thought you meant. Yes, code shouldn't be executed when using `FullLoader`, so that's all good. Admittedly, using this `try: except:` doesn't really enforce proper formatting of the YAML file -- it means its valid _YAML_, but not necessarily that it contains the right kinds of keywords for this particular case. 

I think your solution is good.

hera_sim/simulate.py
steven-murray: I think the way that @r-pascua has done it makes sense (except backwards, see my other comment) -- it may be sometimes desired to overwrite the file parameters with dynamic ones.

hera_sim/simulate.py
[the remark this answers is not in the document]
steven-murray: I'd still suggest doing this.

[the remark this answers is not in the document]
r-pascua: Is it valid to default `**sim_params` as `None`? That is, will the code break if I adjust the function arguments to be `run_sim(self, sim_file=None, **sim_params=None)`?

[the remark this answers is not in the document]
r-pascua: Nevermind; I just tested this out with a simple case and it didn't work. I'll figure out how to rewrite the code appropriately.

hera_sim/simulate.py
```diff
@@ -434,4 +435,75 @@ def add_xtalk(self, model='gen_whitenoise_xtalk', bls=None, **kwargs):
                 vis=self.data.data_array[blt_ind, 0, :, pol_ind],
                 xtalk=xtalk
             )
+    
+    # make a dictionary whose values point to the various methods
+    # used to add different simulation components
+    SIMULATION_COMPONENTS = { 'noiselike_eor':add_eor,
```
r-pascua: I had to move this dictionary to some point *after* the functions being pointed to were defined. If we try to make a pointer to an object which does not exist yet, then python is going to throw an error. In fact, the previous build failed precisely because of this. If you're not sold, then try out this simple code:
```
class TestClass():
    FUNCTIONS = {'func1':func1, 'func2':func2}
    def func1(self):
        return None
    def func2(self):
        return None
```
If you try to run this, you'll get a `NameError`; however, if you move the `FUNCTIONS = ...` line to a point after `func1` and `func2` have been defined, then everything's fine.

hera_sim/simulate.py
```diff
@@ -434,4 +435,80 @@ def add_xtalk(self, model='gen_whitenoise_xtalk', bls=None, **kwargs):
...
+
+            assert isinstance(params, dict), \
+                    'Values of sim_params must be dictionaries. ' + \
+                    "The values for '{}' do not comply.".format(model)
+
+
+        # everything should be in working order at this point, so let's simulate
+        for model, params in sim_params.items():
+            add_component = self.SIMULATION_COMPONENTS[model]
+            if model in uses_no_model:
```
r-pascua: This isn't exactly correct. As of now, there's no guarantee that multiplicative effects will be applied last. @steven-murray, do you know how I can make sure that multiplicative effects are added last?

hera_sim/simulate.py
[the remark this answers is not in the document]
steven-murray: I don't think there should be much you have to do. The only thing would be to change where you check if `sim_params` is `None`: change to `if sim_params` -- if not given, it's an empty dict.

hera_sim/simulate.py
steven-murray: yup, that makes perfect sense. If you really want it at the top (and that _would_ be nice), you can replace the functions/methods with _strings_ giving their names, and then just do `getattr(self, stringname)` when you access it.

hera_sim/simulate.py
steven-murray: oh yes, good catch. I guess you could order the entries in `SIMULATION_COMPONENTS`?

r-pascua: I suppose that could work. I'd have to make some other changes elsewhere, but I think I know what to do.

hera_sim/simulate.py
r-pascua: I like that solution. I implemented it in the latest commit.

hera_sim/simulate.py
```diff
@@ -434,4 +451,69 @@ def add_xtalk(self, model='gen_whitenoise_xtalk', bls=None, **kwargs):
...
+        assert sim_file is None or len(sim_params) is 0, \
+                'Either a simulation configuration file or a dictionary ' + \
+                'of simulation parameters may be passed, but not both. ' + \
+                'Please choose only one of the two to pass as an argument.'
+
+        # if a hook to a simulation file is provided, then read it in
+        if sim_file is not None:
+            with open(sim_file, 'r') as doc:
+                try:
+                    sim_params = yaml.load(doc.read(), Loader=yaml.FullLoader)
```
r-pascua: This is going to be problematic for simulation components that require an interpolation object for the sky temperature model. I can't currently think of a *nice* way of handling these cases. (A straightforward way would be to check for the `Tsky_mdl` key in any of the parameter dictionaries, parse the value, and point to the interpolation object corresponding to it, but this requires the interpolation object to already exist somewhere.)

steven-murray: Hmmm, what exactly is the problem here? Is the problem that it's hard to specify an interpolation object in a YAML file? If you have a dict of available interpolation objects defined somewhere inthe code, you could reference them by name in the YAML file? They definitely have to exist, of course. 

Or is it that you are envisioning that people may create their own interpolation objects outside of `hera_sim` and may want to use them?

hera_sim/simulate.py
```diff
@@ -434,4 +451,68 @@ def add_xtalk(self, model='gen_whitenoise_xtalk', bls=None, **kwargs):
...
+        """
+
+        # keep track of which components don't use models
+        uses_no_model = []
+        for key, val in self.SIMULATION_COMPONENTS.items():
+            func = getattr(self, val)
+            if 'model' not in inspect.signature(func).parameters:
+                uses_no_model.append(key)
+
+        assert sim_file is not None or len(sim_params) is not 0, \
```
steven-murray: use `assert sim_file is not None or sim_params` (non-empty dicts are evaluated as True)

hera_sim/simulate.py
```diff
@@ -434,4 +451,68 @@ def add_xtalk(self, model='gen_whitenoise_xtalk', bls=None, **kwargs):
...
             )
+    
+    
+    def run_sim(self, sim_file=None, **sim_params):
+        """
+        Accept a dictionary or YAML file of simulation parameters and add in
+        all of the desired simulation components to the Simulator object.
+
+        Args:
+            sim_file (str): string providing a hook to a YAML file
```
steven-murray: Add that this is optional.

hera_sim/simulate.py
```diff
@@ -434,4 +451,68 @@ def add_xtalk(self, model='gen_whitenoise_xtalk', bls=None, **kwargs):
...
+            
+            sim_params (dict): dictionary of simulation parameters.
```
steven-murray: Also add that this is optional, *and* tell the user that only one of them can be supplied

hera_sim/simulate.py
r-pascua: The problem here is indeed that it's hard to specify an interpolation object in a YAML file. Another difficulty is that (as far as I know) interpolation objects cannot be saved to disk. (If there is a way to do so that is machine-agnostic, then I'd be happy to be informed of such a way.)

I think that making a dict of available interpolation objects defined somewhere in the code is a quick and easy fix, but I don't think that it's a fix that's made with longevity in mind. I think that it would be better to allow people to make their own interpolation objects outside of `hera_sim` and use them; however, that seems much easier said than done.

There is one obvious way I can think of going about doing this. It would involve creating a helper function `_make_interp_object`, which accepts a hook to a `npz` file which has in it an array of `Tsky` values, as well as arrays of the LSTs and frequencies that the model is evaluated at. We could enforce a regular formatting to this: the frequency array must be saved with the key `'fqs'`; the LST array must be saved with the key `'lsts'`; and the sky temperature model must be saved with the key `'Tsky'`. The helper function would then generate and return an interpolation object using the arrays stored in the `npz` file. I imagine we could also include the ability to choose which type of interpolator to use (but limit it to types supported by `scipy.interpolate`), and allow for users to pass `'h1c'` or `'h2c'` (and so on) for shortcuts to the accepted standard sky temperature model for the given observing season.

Any thoughts on this?

steven-murray: I think your idea is a good one -- though I reckon it's better as a class:

```
class Tsky:
    def __init__(self, datafile, interpolator='linear'):
        self.datafile = datafile
        self.interpolator = interpolator
        # np.load is lazy, so this is fast
        self._data = np.load(self.datafile)
    
    @property
    def freqs(self):
        return self._data['freqs']
    @property
    def tsky(self):
        return self._data['tsky']
    @property
    def lsts(self):
        return self._data['lsts']

    @cached_property
    def _interpolator(self):
       return scipy.interpolate(...)

    def __call__(self, freqs, lsts):
        return self._interpolator(freqs, lsts)
```

Then when loading from YAML you just need to use this class.
            

```

Comment (r-pascua) [reply hera_sim/simulate.py]: Any suggestions of which module to put this class in?

File: hera_sim/simulate.py [hunk 13]
```diff
@@ -434,4 +451,69 @@ def add_xtalk(self, model='gen_whitenoise_xtalk', bls=None, **kwargs):
...
+            One (and *only* one) of the above arguments must be provided. If 
+            *both* sim_file and sim_params are provided, then this function
+            will raise an AssertionError.
+        """
+
+        # keep track of which components don't use models
+        uses_no_model = []
+        for key, val in self.SIMULATION_COMPONENTS.items():
+            func = getattr(self, val)
+            if 'model' not in inspect.signature(func).parameters:
```
r-pascua: This bit of code is breaking in the Python 2.7 build. Apparently the behavior of `inspect.signature` is quite different between the Python 2 and Python 3 implementations. When this line is run in the Python 2.7 build, an error is thrown, stating something along the lines of "modules don't have signatures".

Is it important to fix this to keep Python 2 compatibility? If so, how can we go about fixing it?

steven-murray: Ah yeah, should've thought of that. I'm inclined to say we drop Py2 support -- in <6 months we'll be forced to anyway. So as not to cause a fuss, you can probably just do something like 

```
if sys.version_info < (3, 0):
    <python2 way of doing inspect.signature (look it up)>
else:
   < what you have now>
```

hera_sim/simulate.py
steven-murray: either a new module, or maybe `noise`? I'd go for a new module.

r-pascua: If we put it in a new module, do we expect to add new things to the module other than the `Tsky` class? Any ideas for what to name the module?

hera_sim/simulate.py
r-pascua: sounds good

hera_sim/simulate.py
r-pascua: Another question: do we provide the option for users to choose any interpolator from the `scipy.interpolate` module, and require that they specify which with a string? Or do we restrict it to just a few types of interpolators (e.g. `RectBivariateSpline` and `interp1d`)?

steven-murray: Honestly, I'd keep to a single one, and let the user pass kwargs to that interpolator (like kind='linear'). All the interpolators have different API's. I think it's up to you to choose the one that best suits this problem and just use it.

hera_sim/simulate.py
r-pascua: After a bit of digging around, I've come to the conclusion that there isn't a simple workaround for Py2. The main issue seems to be that `functools.wraps` preserves signatures for python versions 3.4 and later, but it does not for earlier versions of python. So the `_model` decorator overwrites the function signatures when `hera_sim` is used on Py2. It looks like there *is* a solution, but it involves rewriting the `_model` decorator using the `decorator` module.

Is it worth doing all this extra work to support a soon-to-be-deprecated version of python?

steven-murray: Most definitely not. I forgot that it was decorated. You could just raise an error if python 2 is being used for this particular method.

hera_sim/simulate.py
r-pascua: So I've been thinking a bit about this, and the proposed implementation would break the current `hera_sim` API a bit. In particular, `diffuse_foreground` and `thermal_noise` require an interpolation object to be passed, rather than the re-sampled interpolation object. So the proposed method wouldn't jive with what's currently in place.

Here's what I'm thinking in order to solve this problem:
Instead of having a `__call__` method that returns the re-sampled sky temperature model, we implement `noise.resample_Tsky` as a member function of the `Tsky` class. Additionally, we keep important metadata in the `Tsky` class, in particular the units of class properties and the polarizations supported. In order to retrieve the interpolation object, we can implement a `get_interpolator` member function. 

Any thoughts on this? Also, do you have any thoughts on putting the `Tsky` class in the `utils` module?

r-pascua: Another question/comment regarding your suggested format for the `Tsky` class:
What's the motivation behind using 
```
@property
def freqs(self):
    return self._data['freqs']
```
instead of `self.freqs = self._data['freqs']` in the `__init__` method? I understand that this makes it so that a `Tsky` object's frequencies (and so on) can't be changed, since it just pulls them from the loaded `npz` file; however, is this a feature that we want?

steven-murray: I think maybe you mis-understand what `__call__` does? The class that I suggested would act like this:

```
tsky = Tsky("my_data_file.npz")
tsky_at_lsts_and_freqs = tsky(lsts, freqs)

# Can pass directly in as-is
diffuse_foregrounds(lsts, freqs, bl_vec, tsky)
```

i.e. the `__call__` method means you can call the instance itself. In this case, that call will do the resampling, just as a spline object would.

You could put it in `utils` for now -- also import it directly in `__init__` so that if we swap its module it won't hurt too many users.

Concerning the use of `@property`s, I do think we should disallow setting freqs/lsts by the user. If they are changed, the whole object becomes inconsistent. If you want to change them, you should create a new object from scratch. Also, this way the data is lazy-loaded.

r-pascua: Ah, yes, I see what you mean.

 That said, I still think there's some value to having the `Tsky` object instead hold references to the interpolator objects. As a working example, the sky temperature model that's been used in `hera_sim` thus far (saved in `HERA_Tsky_vs_LST.npz`) has sky temperature arrays for two polarizations. If we construct the class so that the resampling method is accessed via `__call__`, then we would need to either have the user specify the polarization used when a `Tsky` object is instantiated, or we would need to enforce that the `npz` files hold data for only one polarization. (We could require that the polarization be specified in the `__call__` method, but that would mess with the API.)

steven-murray: Hmm, that's what I think the `._interpolator` property is. It's a scipy interpolator, sitting in the class waiting to be used.

On the polarization thing: yeah I reckon you pass it at init-time, just like the "interpolation_kind" or whatever. You can imagine that constructing from a YAML is still easy: you specify the datafile and then the other options are strings.

hera_sim/interpolators.py
```diff
@@ -0,0 +1,99 @@
...
+                "key 'lsts'."
+        assert 'tsky' in self._data.keys(), \
+                "The sky temperature array must be saved to the npz file " \
+                "using the key 'tsky'."
+        assert 'meta' in self._data.keys(), \
+                "The npz file must contain a metadata dictionary that can " \
+                "be accessed with the key 'meta'. This dictionary should " \
+                "provide information about the units of the various arrays " \
+                "and the polarizations of the sky temperature array."
+        # now make sure that the sky temperature array has the correct shape
```
r-pascua: @steven-murray do you think it makes sense and is a good idea to require the `npz` files to conform to this structure? That is, does it make sense to require that `npz` files also contain a `meta` dictionary that contains information about the arrays' units and the polarizations included and to require the sky temperature arrays to have `shape==(NPOLS,NLSTS,NFREQS)`?

Also, do you see any clear benefits of implementing `Tsky` more in-line with what you proposed rather than the way I've implemented it here?

hera_sim/data/tests/test_sim_file.yaml
```diff
@@ -1,5 +1,5 @@
 #diffuse_foreground:
-#    Tsky_mdl: HERA_Tsky_mdl['xx']
+#    Tsky_mdl: # XXX how to give a hook to a npz that Travis can find???
```
steven-murray: If you're doing a test, you can maybe write your yaml dynamically as a string, write it to a temp file, and then read it in. That way you know where the npz is. Also remember to delete all test artifacts at the end of your test! Using temporary files is useful for this!

hera_sim/interpolators.py
steven-murray: This looks pretty good. I like the idea of a `meta` dictionary in general, though to be honest, I'd just as well have the `.npz` arrays have an exact specific form here, readable by this specific class, and always interpreted the same way. I think it's overkill to allow for unit conversion etc here at this point. If these `.npz` files were going to be created a lot and by lots of different people, maybe you'd want that. But really, we're going to have a pretty standard/static library of these. That means that `meta` really just comes down to specifying pols. 

Forcing the sky temperature array to have that shape is good. Any well-defined standard is better than none.

hera_sim/interpolators.py
```diff
@@ -0,0 +1,99 @@
...
+
+class Tsky:
+    """
+    # TODO: fill in docstring
+    """
+    def __init__(self, datafile, **interp_kwargs):
+        self.datafile = datafile
+        self._data = np.load(self.datafile, allow_pickle=True)
+        # make sure the correct keys are used
+        assert 'freqs' in self._data.keys(), \
```
steven-murray: Maybe put everything here in a private method `_check_npz_format`.

hera_sim/data/tests/test_sim_file.yaml
r-pascua: Thanks for the advice! I'll rewrite the tests to use temporary files.

hera_sim/interpolators.py
```diff
@@ -0,0 +1,99 @@
...
+        # XXX sky temperature to be shaped incorrectly and pass this assertion
+        assert self.tsky.shape==(len(self.meta['pols']),
+                                 self.lsts.size, self.freqs.size), \
+                "The tsky array is incorrectly shaped. Please ensure that " \
+                "the tsky array has shape (NPOLS, NLSTS, NFREQS)."
+
+        self._interpolators = {}
+
+        # fill in interpolators
+        # XXX do we want to make a hidden member function for generating
```
steven-murray: I think having hidden member properties (actually, `cached_property`s) is a better way to go. This class should be considered a `view` into a specific `npz` file, and should disallow changing the data. There's no reason to ever change the data, which makes it _more clear_ what's going on by making them properties or cached properties.

hera_sim/interpolators.py
```diff
@@ -0,0 +1,99 @@
...
+            # interpolator stability
+            lsts = np.concatenate([self.lsts[-10:]-2*np.pi,
+                                   self.lsts,
+                                   self.lsts[:10]+2*np.pi])
+            tsky_data = np.concatenate([tsky_data[-10:],
+                                        tsky_data,
+                                        tsky_data[:10]])
+            
+            # now make the interpolation object
+            self._interpolators[pol] = RectBivariateSpline(lsts,
```
steven-murray: Also, creating the spline object can take some time. If the user never needs a certain pol, then doing this for every pol is unnecessary. There's two ways around this: the first idea is to pass a non-changeable 'pol' keyword to `__init__`, which would set which polarization this particular instance uses once and for all. This is the easiest way, and is more easily specifiable via YAML. However, it does lack some flexibility, in that once the object is created you can't get at the other polarization.

The other idea would be to make a method:
```
@memoize
def _interpolator(self, pol='xx'):
    j = self.meta['pols'].index(pol)
    < do all the stuff you did to get a single interpolator >
```

Then everything else as you've got it, except to replace calls to `self._interpolators['xx']` with `self._interpolator('xx')`. The `@memoize` function caches results based on the inputs. This means that if 'yy' is never used, it is never calculated, but the 'xx' is only ever calculated once.

hera_sim/interpolators.py
```diff
@@ -0,0 +1,99 @@
...
+            
+            # now make the interpolation object
+            self._interpolators[pol] = RectBivariateSpline(lsts, 
+                                                           self.freqs,
+                                                           tsky_data,
+                                                           **interp_kwargs)
+
+    @property
+    def freqs(self):
+        # XXX units handling?
```
steven-murray: instead of handling units, I'd just put in a good docstring here that tells which units to use. We can maybe add proper units handling later.

hera_sim/interpolators.py
```diff
@@ -0,0 +1,99 @@
...
+        return self._data['lsts']
+    # XXX should we assert that units must be of a certain type, or do we allow
+    # XXX for more flexibility and provide conversions between units?
+
+    @property
+    def meta(self):
+        return self._data['meta'][None][0]
+
+    def get_interpolator(self, pol='xx'):
+        self._check_pol(pol)
```
steven-murray: this would not be required any more.

hera_sim/interpolators.py
```diff
@@ -0,0 +1,99 @@
...
+        return self._interpolators[pol]
+
+    def resample_Tsky(self, lsts, freqs, pol='xx'):
```
steven-murray: I still think using the `__call__` syntax is nicer here. Ultimately this _is_ an interpolator object. That to me suggests you should be able to "call" it on your new points.

hera_sim/simulate.py
steven-murray: This looks good

hera_sim/tests/.coverage
```diff
@@ -0,0 +1 @@
...1,72,73,74,75,76,77,78,80,82,83,84,85,86,87,89,91,92,93,94,96,98,99,100,101,102,103,104,105,106,107,108,109,110,113,114,115,116,118,119,120,121,122,124,125,126,127,129,131,132,134,138,140,142,143,144,145,146,147,148,149,150,151,152,153,154,155,163,165,195,197,198,200,202,204,206,208,210,212,213,214,215,216,217,218,220,221,222,223,225,227,228,230,236,238,239,241,242,244,246,249,250,251,252,254,255,256,257,258,260,268,269,270,271,272,273,275,277,279,280,282,285,286,287,288,289,290,292,305,307,308,309,310,311,312,313,315,328,330,331,332,333,334,335,336,338,350,351,352,353,354,355,356,357,359,371,372,373,374,375,376,377,378,379,380,381,382,384,397,398,399,400,401,402,403,404,405,406,407,408,409,410,411,417,418,419,420,421,423,424,425,426,427,428,429,430,431,432,433,439,440,441,442,443,444,445,446,450,451,452,453,454,456,458,476,477,478,479,480,482,483,485,486,488,489,492,493,494,495,496,497,498,499,500,501,502,503,504,505,506,507,508,509,511,512,513,514,515,516,517,518,519,520,526,527,529],"/home/bobby/HERA/dev/make_data/hera_sim/hera_sim/interpolators.py":[3,5,6,8,11,12,13,14,16,17,20,24,27,37,38,42,47,48,52,53,54,55,56,57,60,61,62,63,65,68,70,73,75,78,82,84,86,87,88,90,91,92,94,95]}}
```
steven-murray: put this file in .gitignore!

hera_sim/simulate.py
```diff
@@ -434,4 +454,93 @@ def add_xtalk(self, model='gen_whitenoise_xtalk', bls=None, **kwargs):
...
+        """
+
+        # keep track of which components don't use models
+        uses_no_model = []
+        for key, val in self.SIMULATION_COMPONENTS.items():
+            func = getattr(self, val)
+            # raise a NotImplementedError if using Py2
+            if sys.version_info.major < 3 or \
+               sys.version_info.major > 3 and sys.version_info.minor < 4:
+                raise VersionError("Please use a version of Python >= 3.4.")
```
r-pascua: Do you think it's better to drop Py2 support entirely, or just manually define the entries in `uses_no_model` in the case that someone is running a version < 3.4?

hera_sim/simulate.py
```diff
@@ -493,6 +500,26 @@ def run_sim(self, sim_file=None, **sim_params):
...
+                    # since we're loading in a YAML file, we need to fix the
+                    # parameters for any components that require a Tsky model
+                    for component, params in sim_params.items():
+                        if "Tsky_mdl" in params.keys():
+                            # given the behavior of the interpolators.Tsky
+                            # class, we need a dictionary with a 'file' key
+                            # and a 'pol' key
+                            # XXX think about the best way to assert this
+                            # XXX do we assert that it has the correct keys?
+                            assert isinstance(params["Tsky_mdl"], dict), \
```
steven-murray: again, probably easier and more error-proof to use duck-typing, eg:

```
try:
    tsky = Tsky(params['Tsky_mdl']['file'], pol=params['Tsky_mdl'].get("pol", 'xx'), **params['Tsky_mdl'].get("interp_kwargs", {}))
except KeyError:
    raise KeyError("please ensure that tsky_mdl has a file and pol key" )
except TypeError:
    raise TypeError("please ensure that tsky_mdl is a dict")

File: hera_sim/simulate.py (hunk 23 above)
Comment (steven-murray) [reply hera_sim/simulate.py]: Nope, just drop it entirely for this function, I reckon.

**Comment by r-pascua** [comment]
Awesome! However, last night I realized that there is still one issue that needs to be fixed... I'll comment on the line of code.

**Comment by r-pascua** [comment]
> Awesome! However, last night I realized that there is still one issue that needs to be fixed... I'll comment on the line of code.

Nevermind! I thought there was going to be an issue with `gen_cross_coupling_xtalk`, but I checked the `add_xtalk` method in the `Simulator` class, and you already handled the issue I thought it was going to have!

File: hera_sim/tests/test_interpolators.py [hunk 25]
```diff
@@ -0,0 +1,59 @@
...
+    # make arrays for npz files
+    lsts = np.linspace(0,np.pi,10)
+    freqs = np.linspace(0.1,0.2,50)
+    ok_tsky = np.array([np.ones((lsts.size,freqs.size))])
+    bad_tsky = np.array([np.arange(15).reshape(3,5)])
+    # make metadata dict
+    pols = ('xx',)
+    meta = {'pols':pols}
+    # start saving files
+    np.savez(temp_dir+'/no_freqs', lsts=lsts, tsky=ok_tsky, meta=meta)
```
steven-murray: prefer `os.path.join()` instead of adding `/xxx` to filename

hera_sim/tests/test_interpolators.py
```diff
@@ -0,0 +1,59 @@
...
+import tempfile
+import shutil
+
+from scipy.interpolate import RectBivariateSpline
+from hera_sim.interpolators import Tsky
+from hera_sim.data import DATA_PATH
+from nose.tools import raises
+
+def test_interpolator():
+    # make a temporary file
```
steven-murray: You don't seem to be actually using the tempfile module in this function?

hera_sim/tests/test_simulator.py
```diff
@@ -155,3 +157,139 @@ def test_adding_vis_but_also_returning():
...
+                rfi_scatter: 
+                    chance: !!float 0.99
+                    strength: !!float 5.7
+                    std: !!float 2.2
+                    """.format(DATA_PATH, DATA_PATH))
+        sim = create_sim()
+        sim.run_sim(tmp_sim_file)
+        assert not np.all(np.isclose(sim.data.data_array, 0))
+        # delete the temp file
+        remove(tmp_sim_file)
```
steven-murray: I don't think you should have to manually remove a temp file?

hera_sim/tests/test_simulator.py
```diff
@@ -155,3 +157,139 @@ def test_adding_vis_but_also_returning():
...
+        with open(tmp_sim_file, 'w') as sim_file:
+            sim_file.write("""
+                pntsrc_foreground:
+                    nsrcs: 5000
+                    """)
+        sim_params = {"diffuse_foreground": {"Tsky_mdl":HERA_Tsky_mdl['xx']} }
+        sim = create_sim()
+        sim.run_sim(tmp_sim_file, **sim_params)
+        # remove test file
+        remove(tmp_sim_file)
```
steven-murray: again, removing tmp file

hera_sim/tests/test_simulator.py
```diff
@@ -155,3 +157,139 @@ def test_adding_vis_but_also_returning():
...
+        with open(tmp_sim_file, 'w') as sim_file:
+            sim_file.write("""
+                this:
+                    is: a
+                     bad: file
+                     """)
+        sim = create_sim()
+        sim.run_sim(tmp_sim_file)
+        # delete the file
+        remove(tmp_sim_file)
```
steven-murray: again, removing tmp file

hera_sim/tests/test_simulator.py
```diff
@@ -155,3 +157,139 @@ def test_adding_vis_but_also_returning():
...
+        with open(tmp_sim_file, 'w') as sim_file:
+            sim_file.write("""
+                diffuse_foreground: 
+                    Tsky_mdl: 
+                        pol: xx
+                        """)
+        sim = create_sim()
+        sim.run_sim(tmp_sim_file)
+        # delete the file
+        remove(tmp_sim_file)
```
steven-murray: again, removing tmp file

hera_sim/tests/test_simulator.py
```diff
@@ -155,3 +157,139 @@ def test_adding_vis_but_also_returning():
...
+        tmp_sim_file = tempfile.mkstemp()[1]
+        with open(tmp_sim_file, 'w') as sim_file:
+            sim_file.write("""
+                diffuse_foreground:
+                    Tsky_mdl: 13
+                    """)
+        sim = create_sim()
+        sim.run_sim(tmp_sim_file)
+        # delete the file
+        remove(tmp_sim_file)
```
steven-murray: again removing tmp file

hera_sim/tests/test_simulator.py
r-pascua: For these tests, I created the temporary file using `tempfile.mkstemp`, which requires the user to handle data cleanup, so that files need to be manually removed. I can figure out how to work with the `tempfile.NamedTemporaryFile` class so that `tempfile` handles cleanup automatically if that's what you'd prefer.

hera_sim/tests/test_interpolators.py
r-pascua: will adjust accordingly

hera_sim/tests/test_interpolators.py
r-pascua: Nope; I only need a temporary `.npz` file for this test. So I thought it would be best to generate some data in the test, save it to a `.npz`, use the `.npz` to instantiate a `Tsky` object, then delete the `.npz` when I'm done with it.

hera_sim/tests/test_simulator.py
steven-murray: Nope, all good.

hera_sim/tests/test_interpolators.py
steven-murray: You can definitely save `.npz` files to the temp dir though?

hera_sim/rfi.py
```diff
@@ -264,14 +264,17 @@ def rfi_dtv(fqs, lsts, rfi=None, freq_min=.174, freq_max=.214, width=0.008,
...
+            fq_ind_min = np.argwhere(band <= fqs)[0][0]
+            fq_ind_max = fq_ind_min + int(width / delta_f) + 1
+            this_rfi = rfi[:, fq_ind_min:min(fq_ind_max, fqs.size)]
+
+            rfis = np.random.uniform(size=lsts.size) <= chnc
+            this_rfi[rfis] += np.atleast_2d(np.random.normal(strngth, str_std, size=np.sum(rfis)) * np.exp(
+                2 * np.pi * 1j * np.random.uniform(size=np.sum(rfis))
+            )).T
+        except IndexError:
+            pass
```
steven-murray: Is it reasonable to always just pass silently here? Shouldn't we be filling in the part of the band that is DTV if it overlaps with an edge?

steven-murray:
Can you also move the notebook into the `docs/` folder, and add an entry so that it shows up in the docs? Also, the YAML file doesn't really need all the `!!float` stuff. You could probably clean it up a little.

r-pascua:
> Can you also move the notebook into the `docs/` folder, and add an entry so that it shows up in the docs? Also, the YAML file doesn't really need all the `!!float` stuff. You could probably clean it up a little.

I'll get on that.

hera_sim/rfi.py
r-pascua: That's a good point. I figure what we could do, then is modify the line `fq_ind_min = np.argwhere(band <= fqs)[0][0]` to instead be `fq_ind_min = np.argwhere(band + width <= fqs)[0][0]`, but keep the `try: except:` clause in there to catch the case where `np.argwhere` returns an empty list. Does that make sense?

r-pascua: I actually think that might have some unwanted behavior... I'll have to think about this a little more carefully.

r-pascua: After reading the documentation and thinking about it, I think that it *is* reasonable to pass silently here. The `IndexError` is raised only in the case that `np.all(fqs < band)` is `True`. Since `band` is the lower end of the DTV channel, if an `IndexError` is raised, then all of the frequencies in that DTV channel are strictly greater than all of the frequencies passed to `rfi_dtv`.

I think that the developer of this code figured that the following clause would ensure that this never happens:
```
if fqs.max() <= bands.max():
    bands = bands[:-1]
```
This works fine under the assumption that there is at least one frequency in `fqs` that is greater than `bands.min()`, but it breaks for the case that `fqs.max() < bands.min()` is `True`.

r-pascua: A quick follow-up to clarify slightly: if `fqs.max() < bands.min()`, then `np.argwhere(band <= fqs)[0][0]` will raise an `IndexError` for every `band` in `bands`; however, if `fqs.max() > bands.min()`, then the line `if fqs.max() <= bands.max(): bands = bands[:-1]` should ensure that `np.argwhere(band <= fqs)` will never return an empty array.

steven-murray: LOL, I was the developer of this code... I can't remember what I did or why I did it!

Having a look now, you probably just want to do something like:

```
bands = bands[np.logical_and(bands >= fqs.min(), bands < fqs.max())]
```

just after creating the `bands` array. This will ensure all bands are within the freq range.

r-pascua: Ok, I agree that this will ensure that all bands are within the freq range; however, if the simulated frequencies are all below the lower-edge of the DTV band (which is the case for the `Simulator` object made from `create_sim` in the `test_simulator` module), then this will return an empty array, which will cause `np.argwhere(band <= fqs)[0][0]` to throw an `IndexError`.

r-pascua: So I think it makes sense to keep the `try: except:` clause, but I think it may be a good idea to add a warning notifying the user that their simulated frequencies are outside of the DTV channels.

steven-murray: I don't think this will throw an error: if `bands` is empty, then the loop won't be entered at all.

One case that's not quite covered by my suggestion is where `freq_min` is smaller than `fqs.min()`, but only just, so that part of that DTV band overlaps with the lowest channels. With my suggestion, this band gets cut entirely. You'll probably have to special-case that. 

I'm up for still using the try: except: anyway, in case we botch something. Or perhaps it's better to use try: except: than limiting the `bands` array: scroll through all bands, and if its close enough or within `fqs`, do what is necessary, and otherwise just warn the user -- it's likely that if this happens they have done something wrong themselves.

In this case, you have to special case where `band` is less than `fqs.min() - width`, and warn the user. Also special-case where `band` is between `fqs.min() - width` and `fqs.min()`, and use the first few indices.

hera_sim/tests/test_interpolators.py
steven-murray: Did you end up doing the `.npz` in a temp file? I'm fine with either way, so I guess it doesn't matter :-)

hera_sim/rfi.py
r-pascua: Hmm. Should I just raise an issue on this and work on figuring it out there? This *is* a separate problem from the actual thing I'm trying to get pulled in here.

hera_sim/tests/test_interpolators.py
r-pascua: I made a temporary directory with `tempfile.mkdtemp` and saved the `.npz` file there.

hera_sim/rfi.py
steven-murray: Yes, that would be great. However, the tests have to all work for it to be merged.

r-pascua: So it seems like there are two ways for me to make sure the tests will work. The first is to just not try to add RFI from DTV when testing `run_sim`. The second is to leave in the `try: except:` clause, and then either include or exclude DTV from the `run_sim` test. Either way, `rfi_dtv` can't really be tested in the `test_simulator` module, since the `create_sim` function initializes a `Simulator` object whose frequencies range from 100 MHz to (100 + 1000/1024) MHz, which is well below the DTV range.

steven-murray: Yup, just leave it out of the tests for now, I reckon

r-pascua: Cool. All the tests are passing. Can I finally get this thing merged in?

hera_sim/rfi.py
```diff
@@ -274,6 +274,8 @@ def rfi_dtv(fqs, lsts, rfi=None, freq_min=.174, freq_max=.214, width=0.008,
                 2 * np.pi * 1j * np.random.uniform(size=np.sum(rfis))
             )).T
         except IndexError:
+            # this is only raised if fqs.max() < bands.min(), which
```
steven-murray: Remember to raise an issue about this

r-pascua: Will do!

Issue #25, opened 2019-01-29
homogenization of model functions


steven-murray:
Some of the model functions throughout the code act differently to the others (eg. you have to call one function to get a result and pass that to another, to achieve a single effect). I think this should be addressed to make things more uniform.

steven-murray:
This ties in with some discussion we've been having on #46, in which we've discussed being able to automatically find all the model functions that have been defined. 

I propose we could use a "plugin" API for this. In short, for each _kind_ of model, we define an abstract base class, which refers to a metaclass which keeps track of implemented actual models. Once `hera_sim` has been loaded, these are automatically discoverable, _and_ have the benefit of being able to enforce a uniform API.

A brief example of how this can be done:

```
# plugins.py
def plugin_mount_factory(abstract=False):
    class PluginMount(type):
        def __init__(cls, name, bases, attrs):
            if not hasattr(cls, '_plugins'):
                cls._plugins = {}
            elif abstract:
                cls._plugins[cls.__name__] = cls

        def _get_plugin(cls, *args, **kwargs):
            return cls._plugins[cls.__name__](*args, **kwargs)

    return PluginMount

# model.py
class ModelBase(metaclass=plugin_mount_factory(abstract=True)):
    def __init__(self, arg1, arg2,...):
        self.arg1 = arg1
        self.arg2 = arg2

    def calculate(self, a, b, c, ...):
        pass

class Model1(ModelBase):
    def calculate(self, a, b, c):
        ....
        return calculate_model
```

Notably, each specific model should *not* implement a new `__init__`, and should conform to the same input parameters for `calculate`. 

A list of all plugins for this model is gotten by `ModelBase._plugins.values()`, and their names as `ModelBase._plugins.keys()`.

Issue #24, opened 2019-01-29
hera specific variables lying around the codebase


steven-murray:
There are a lot of attributes of HERA lying around the codebase (eg. HERA_LOT_LON_ALT). I think it makes more sense to compile these kinds of attributes into a config file (YAML? PY?) which can be accessed directly. All functions in ``hera_sim`` should *by default* use this HERA config file, but can be set to another file (and each parameter can be individually overwritten in the function call). I'm assuming there's already a kind of file format for this kind of thing, but if not we should make it.

PR #47, opened 2019-07-22
H2c sim

r-pascua:
This is a first attempt at doing the `hera_sim` rewrite. There is some repeated code, but that's mainly a consequence of the initial handling of `resample_Tsky` being somewhat messy (generating the interpolation object at the beginning of the module script), and seems kind of necessary in order to allow for more flexibility in the future. 

I've been thinking about doing a somewhat bigger overhaul, perhaps including the changes suggested in issue #24 if it is possible to do so in a somewhat natural way with the spirit of the rewrite. I anticipate pushback due to changes to the API, but PR #46 seems to have revived some interest in addressing issue #25, which would bring about a substantial change to the API. If there is substantial interest in addressing issue #25, then it may be worth the effort to do a bigger overhaul that makes the simulation function calls conform to a more standard/general format, depending on what sort of a timeline the season-switching functionality is on.


r-pascua:
I've noticed that there are quite a few errors regarding module testing. I'll work on updating the tests so these errors go away (so far it seems like a consequence of having moved stuff around).

PR #48, opened 2019-07-25
Move test of sim_red_data over from hera_cal.redcal

jsdillon:
This will help us finally close https://github.com/HERA-Team/hera_cal/issues/379

### DIFF
--- a/hera_sim/tests/test_vis.py
+++ b/hera_sim/tests/test_vis.py
@@ -94,6 +94,111 @@ class TestVisCpu(unittest.TestCase):
             v[:, 0, 1], 1 + np.exp(-2j * np.pi * np.sqrt(0.5)), 7
         )
 
+class TestSimRedData(unittest.TestCase):
+
+    def test_sim_red_data(self):
+        #antpos = build_linear_array(5)
+        #reds = om.get_reds(antpos, pols=['xx'], pol_mode='1pol')
+        reds = [[(0, 1, 'xx'), (1, 2, 'xx'), (2, 3, 'xx'), (3, 4, 'xx')],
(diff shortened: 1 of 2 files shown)


hera_sim/vis.py
```diff
@@ -162,7 +163,7 @@ def hmap_to_I(h):
 #     return gains, true_vis, data
 
 
-# XXX from hera_cal.redcal
+# from hera_cal.redcal
 def sim_red_data(reds, gains=None, shape=(10, 10), gain_scatter=0.1):
     """
     Simulate noise-free random but redundant (up to differing gains) visibilities.
```
steven-murray: Are these noise-free? I see you call `noise.white_noise` in the body of the function.

hera_sim/tests/test_vis.py
```diff
@@ -94,6 +94,111 @@ def test_values(self):
             v[:, 0, 1], 1 + np.exp(-2j * np.pi * np.sqrt(0.5)), 7
         )
 
+class TestSimRedData(unittest.TestCase):
```
steven-murray: Other than knowing that this is testing `sim_red_data`, I have no idea what you are testing here. Could you supply a few higher-level comments?

hera_sim/vis.py
jsdillon: Yes. The visibilities and gains are random, but the baselines that should be redundant are perfectly redundant. I'll clarify that it's thermal noise free.

hera_sim/tests/test_vis.py
jsdillon: Done.

jsdillon:
@steven-murray I think this is ready for your review again.

PR #49, opened 2019-07-29
Bring sim_red_data up to date with changes that had been made on her_cal

jsdillon:
In particular, the code was copied over before this PR, which made tests non-deterministic:

https://github.com/HERA-Team/hera_cal/pull/456

### DIFF
--- a/hera_sim/vis.py
+++ b/hera_sim/vis.py
@@ -181,32 +181,20 @@ def sim_red_data(reds, gains=None, shape=(10, 10), gain_scatter=0.1):
         dict: true underlying visibilities in the {(ind1,ind2,pol): np.array} format
         dict: simulated visibilities in the {(ind1,ind2,pol): np.array} format
     """
-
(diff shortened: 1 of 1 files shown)


hera_sim/vis.py
```diff
@@ -181,32 +181,20 @@ def sim_red_data(reds, gains=None, shape=(10, 10), gain_scatter=0.1):
...
-            [
-                ant
-                for bls in reds
-                for bl in bls
-                for ant in [(bl[0], jnum2str(jstr2num(bl[2][0]))), 
-                            (bl[1], jnum2str(jstr2num(bl[2][1])))]
-            ]
-        )
-    )
+    ants = sorted(list(set([ant for bls in reds for bl in bls for ant in
```
nkern: sorted automatically takes a set into a list, so technically I think the `list` func is unnecessary, but not a big deal to me honestly

PR #52, opened 2019-08-06
Updated run_sim to allow for flexible ordering of simulation components

r-pascua:
Note that the order in which simulation components are added is determined by the order in which they are defined in the configuration YAML or the configuration dictionary.


r-pascua:
> This looks good (once tests are passing). One extra comment: the docstring for `sim_params` is slightly wrong, since it is now an unpacked dictionary (or a bunch of keywords).

Any suggestions on how to update the docstring? I'm a little fuzzy on the precise language to use here, and I also want the description to be clear to the user.

steven-murray:
Yeah, maybe something like

```
**sim_params : simulation parameters
    Each param must take the form `model = {param_name: param, ...}`, 
    where `model` specifies the simulation component to be added,
    and the dictionary provides all the keyword arguments and their 
    desired values for that particular component.
```

r-pascua:
Done. I've realized that the current implementation does not allow the user to use custom `omega_p` values (along with some other model kwargs that are typically fed in as `numpy` arrays); however, `Tsky_mdl` is the only one of these that is *mandatory* for some of the simulation components. This is actually the biggest issue I've had with developing the default setting stuff in the `h2c_sim_v2` branch. I will work on creating a unified scheme for dealing with this once I get a working version of `h2c_sim_v2` up and we've agreed on how to implement the tools developed in that branch.

steven-murray:
Obviously these things are easy to pass dynamically, but from YAML, you could try defining a new Loader class that adds a tag such as !ndarray and knows how to read in .npz/.npy arrays. I'm gonna merge this.

PR #53, opened 2019-08-08
Rfi dtv update

r-pascua:
Forgot to submit this PR the other day, but I think this should do what we want it to do.

### DIFF
--- a/hera_sim/rfi.py
+++ b/hera_sim/rfi.py
@@ -238,10 +238,8 @@ def rfi_dtv(fqs, lsts, rfi=None, freq_min=.174, freq_max=.214, width=0.008,
 
     bands = np.arange(freq_min, freq_max, width)  # lower freq of each potential DTV band
 
-    # If the bands fit exactly into freqs, the upper band will be the top freq
-    # and we need to ignore it.
-    if fqs.max() <= bands.max():
-        bands = bands[:-1]
+    # ensure that only the DTV bands which overlap with the passed frequencies are kept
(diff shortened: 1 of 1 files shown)


hera_sim/rfi.py
```diff
@@ -264,19 +267,13 @@ def rfi_dtv(fqs, lsts, rfi=None, freq_min=.174, freq_max=.214, width=0.008,
...
-            rfis = np.random.uniform(size=lsts.size) <= chnc
-            this_rfi[rfis] += np.atleast_2d(np.random.normal(strngth, str_std, size=np.sum(rfis)) * np.exp(
-                2 * np.pi * 1j * np.random.uniform(size=np.sum(rfis))
-            )).T
-        except IndexError:
-            # this is only raised if fqs.max() < bands.min(), which
-            # would mean that DTV wouldn't be observed in any channel
-            pass
+        fq_ind_min = np.argwhere(band <= fqs)[0][0]
+        fq_ind_max = fq_ind_min + int(width / delta_f) + 1
```
steven-murray: this line will not be correct if `band < fqs.min()`. In that case, the width should be reduced because only part of the width is overlapping. To get `fq_ind_max` you could do something like:

```
try:
    fq_ind_max = np.argwhere(band + width <=fqs)[0][0]
except IndexError: # the band goes over the upper edge
    fq_ind_max = len(fqs) - 1
```

r-pascua: Done.

Issue #23, opened 2019-01-29
common array layouts specified by functions


steven-murray:
When creating empty data using ``io.empty_uvdata``, it might be useful to be able to specify some common antenna layouts just with a function call (like, ``hexagon(nside, spacing=14)``), rather than having to go write them down yourself. I'm proposing a simple module for functions that return these kinds of algorithmic layouts.

steven-murray:
This will be fixed as part of #24 and #54

PR #54, opened 2019-08-08
H2c sim v2

r-pascua:
New additions this PR:
- new `defaults` module
  - `defaults._Defaults` class:
    - Provides an interface for managing default settings of various functions within `hera_sim`.
    - Provides keywords for season defaults: `'h1c'`, `'h2c'`
  - `defaults.defaults`: singleton instance of :class: `_Defaults`
    - Default settings chosen via `defaults.set(path_to_default_config or season_default_keyword)`
  - `defaults._defaults`: function used to decorate various functions in `hera_sim`
    - Decorator acts only to set default settings of function it's applied to
- new `__yaml_constructors` module
  - Module for defining new YAML tags for generating `hera_sim.interpolators` interpolation objects
  - Not intended to be interfaced with by end-user
- new `config/` directory
  - Directory for storing configuration YAML files

Modifications to existing structures in this PR:
- new `interpolators` module:
  - `interpolators.freq_interp1d` class
    - Class for generating interpolation objects which depend only on frequency
    - Interpolation objects generated from `.npy` or `.npz` files, depending on interpolator type
    - Currently supported interpolator types: `numpy.poly1d`, `scipy.interpolate.interp1d`
    - See documentation for details
- new `data/` directory
  - Various `.npy` and `.npz` files which may be used to reconstruct various HERA-specific objects
    - e.g. `HERA_H1C_BEAM_POLY.npy` stores polynomial coefficients characteristic of HERA beam (as a function of frequency) for the H1C observing season.


hera_sim/noise.py
```diff
@@ -7,24 +7,26 @@
...
+    replaced in the future.
+    """
+    datafile = _check_path(datafile)
+    return np.load(datafile)
 
-HERA_BEAM_POLY = np.array([8.07774113e+08, -1.02194430e+09,
-                           5.59397878e+08, -1.72970713e+08, 3.30317669e+07, -3.98798031e+06,
-                           2.97189690e+05, -1.24980700e+04, 2.27220000e+02])  # See HERA Memo #27
+# XXX I don't like this. Also figure out what to do about omega_p.
+HERA_BEAM_POLY = _get_hera_bm_poly()
```
r-pascua: It turns out this doesn't work the way I'd like it to--`HERA_BEAM_POLY` just gets defined at runtime and then is insensitive to changing defaults.

steven-murray:
This is a really cool idea. I think we should try to be pretty specific about the scope of this PR. Some things I think it should do (sorry I don't have a lot of time to think right now so these may be a bit scattered):

1. Enable simple determination of what I'll call `model`s that belong to what I'll call `component`s. That is, say "gains" are a `component`, then there may be several `model`s that apply that `component` in different ways. However this is defined, each `model` that is defined should be easy to find.
2. Each `component` should have a consistent interface, so that all models can be called in the same simple way.
3. Users should be able to define their own `model`s for existing `component`s without touching the source code. Not quite sure if they should be able to (easily) define new components.
4. For _all_ components/models, any argument for which reasonable defaults can be applied directly from HERA definitions, should be that default (this is `hera_sim` after all). 
5. To enable quickly switching between seasons (and ostensibly instruments altogether), these defaults should be collected in some kind of structure, and the collection of these structures should be able to be searched by keyword (eg `"h1c"`). 
6. _At runtime_, in an interactive session, there should be a call that goes something like `hera_sim.set_defaults('h2c')` that instantly sets all future defaults to that particular set of defaults. 
7. Users should be able to specify their own set of defaults without touching the code.
8. Defaults should be over-rideable when calling functions, just like normal defaults.
9. The `type` and range of default parameters should be well-defined, not loosely defined (so that badly defined parameters can be identified).
10. There should be scope for updating the range of default parameters in the normal sense of updating code (i.e. no new backwards incompatible changes except for major versions).
11. Preferrably, whichever way we implement the defaults, there will be some mapping of this to a non-python config file (eg. YAML), so that these things can be passed around easily.

Anyway, I have some ideas on how to achieve these, but can't comment on that right now. Would you like to add to the list, and maybe confirm which of these the current code already supports?

r-pascua:
Thanks for these ideas! I'll comment on each idea separately.

1. I like the sound of this. I imagine it will involve rewriting much of `hera_sim` in a more OOP manner, akin to what was suggested in Issue #25. This could also help us update some of the changes introduced in PR #46 so that they are automated (e.g. constructing the `SIMULATION_COMPONENTS` dictionary).
2. Agreed, but sounds like an extra addition to (1).
3. Sounds like we might want to do this with a decorator `@component`?
4. Interesting. Not exactly sure how I'd go about implementing this, but certainly something to keep in mind.
5. This is already implemented.
6. This is already implemented. The current version uses `h1c` as the default, but it also doesn't immediately activate the defaults: the initial behavior is to use the defaults defined in the function signatures, but this may be overridden by executing `hera_sim.defaults.activate_defaults()`. This behavior can be reverted with `hera_sim.defaults.deactivate_defaults()`.
7. This is something I intend to implement, but have not yet implemented.
8. This is already implemented.
9, 10. I'm not sure what you're referring to when you say "range of default parameters".
11. This is already implemented. I created a `config` directory in `hera_sim` and currently have YAMLs for `h1c` and `h2c` stored therein.

Additional suggestions/notes:
12. The repository has been updated so that instrument/season-specific variables are now stored in either `npy` or `npz` files. `npz` files are intended to be used to make interpolation objects from `scipy.interpolate` and thus are required to follow a particular formatting convention. All of the new `npz` files in the `data` directory are formatted appropriately. Season-specific variables that were previously defined in code have not been removed, but are now defined by loading in the appropriate file from `data`. I would like to eventually remove these from the code base (but still have a method for retrieving them, possibly accessed via the `defaults` module).
13. Some functions should be modified so that any season-specific variables they reference are part of their kwargs. A particular example of something that does not adhere to this is `gen_bandpass`. These season-specific variables tend to be somewhat complex in nature (either arrays or interpolation objects), so we need to develop a standard of handling these. I am envisioning that this will be done using the tools in `interpolators` as well as the `@_defaults` decorator I've developed.

hera_sim/interpolators.py
```diff
@@ -131,3 +157,106 @@ def _check_pol(self, pol):
                 "The metadata contains the following polarizations: " \
                 "{}".format(self.meta['pols'])
 
+# XXX can we think of a better name for this? it's a bit unfortunate that
+# XXX interp1d is already taken...
+class Hera1dInterp:
```
nkern: freq_1dinterp something like that perhaps

hera_sim/defaults.py
```diff
@@ -0,0 +1,185 @@
...
+                          "then please use a version of Python newer than 3.4.")
+            return False
+        else:
+            return True
+
+    def _handler(self, func, *args, **kwargs):
+        if self._version_is_compatible:
+            @functools.wraps(func)
+            def new_func(*args, **kwargs):
+            # XXX do we want to use the season defaults by default?
```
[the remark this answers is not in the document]
steven-murray: Thinking about this a bit more, perhaps it would be good to have a standalone function, say `set_defaults(config)`, which the user can call. Internally, it just does `_defaults._get_config(config)` then `activate_defaults`. I think whenever the user actually manually _sets_ defaults, the behaviour should be to use them everywhere, rather than func sig.

hera_sim/defaults.py
```diff
@@ -0,0 +1,185 @@
...
+
+                return func(*args, **new_kwargs)
+        else:
+            @functools.wraps(func)
+            def new_func(*args, **kwargs):
+                return func(*args, **kwargs)
+        
+        return new_func
+
+    def _retrieve_models(self, defaults):
```
steven-murray: I don't think this is the right place for this function. On face value, it belongs to the interpolator class.

hera_sim/interpolators.py
```diff
@@ -4,7 +4,26 @@
...
-from scipy.interpolate import RectBivariateSpline
+from scipy.interpolate import RectBivariateSpline, interp1d
+from hera_sim.data import DATA_PATH
+from os import path
+
+INTERP_OBJECTS = {"1d": ("beam", "bandpass"),
+                  "2d": ("Tsky_mdl", ) }
+
+def _check_path(datafile):
+    # if the datafile is not an absolute path, assume it's in the data folder
```
steven-murray: I'm not sure this is a good assumption. Least surprise says that the path is the user's current path.

hera_sim/interpolators.py
```diff
@@ -4,7 +4,26 @@
...
+    if datafile[0] != "/":
```
steven-murray: I believe there's an `os.path` function for checking if a path is absolute (in an OS-independent way)

hera_sim/interpolators.py
```diff
@@ -68,6 +88,12 @@ def __init__(self, datafile, pol='xx', **interp_kwargs):
     def __call__(self, lsts, freqs):
         return self._interpolator(lsts, freqs)
 
+    # XXX is there a cleaner way of doing this? (I want to be able to find out
```
steven-murray: I'm sure there is. What is the purpose of this? It seems like it should happen the other way round: the parameter that accepts it should check whether what is passed is indeed an instance of `Tsky`.

hera_sim/interpolators.py
```diff
@@ -68,6 +88,12 @@ def __init__(self, datafile, pol='xx', **interp_kwargs):
...
+    # XXX what this interpolator is used for without making an instance of it)
```
steven-murray: If you want exactly this behaviour, you're better off just defining `_names` as a class attribute:

```
class Tsky:
    _names = ("Tsky_mdl", )
    ...
```

steven-murray: but that is _not_ good encapsulation

hera_sim/interpolators.py
```diff
@@ -4,7 +4,26 @@
...
+    assert path.exists(datafile), \
+            "If datafile is not an absolute path, then it is assumed to " \
+            "exist in the hera_sim.data folder. The datafile passed could " \
+            "not be found; please ensure that the path to the file exists"
+    return datafile
+
+def _read_npy(npy):
+    return np.load(_check_path(npy))
 
 class Tsky:
```
steven-murray: Since this is primarily for easily defining the Tsky from YAML, why not go the extra mile and make it a new YAML tag: https://stackoverflow.com/a/43060743/1467820? That way you should be able to define Tsky in any YAML by

```
tsky: !Tsky
    datafile: 'path/to/file.npz'
    pol: 'xx'
```

steven-murray: It will automatically be loaded by `yaml.load` as an instance of `Tsky`.

hera_sim/interpolators.py
```diff
@@ -131,3 +157,106 @@ def _check_pol(self, pol):
...
+                "Interpolator choice must either be 'poly1d' or 'interp1d'."
+
+        if self._interp_type=='interp1d':
+            assert self.datafile[-4:]=='.npz', \
+                    "In order to use an 'interp1d' object, the reference file " \
+                    "must be a '.npz' file."
+            assert self._obj in self._data.keys() and 'freqs' in self._data.keys(), \
+                    "You've chosen to use an interp1d object for modeling the " \
+                    "{}. Please ensure that the `.npz` archive has the following " \
+                    "keys: 'freqs', '{}'".format(self._obj, self._obj)
```
steven-murray: is the last key here supposed to just be "object" rather than `self._obj`?

hera_sim/interpolators.py
```diff
@@ -131,3 +157,106 @@ def _check_pol(self, pol):
...
+            obj = self._data[self._obj]
+            freqs = self._data['freqs']
+
+            # use a cubic spline by default, but override this if the user
+            # specifies a different kind of interpolator
+            kind = self._interp_kwargs.pop('kind', 'cubic')
+            return interp1d(freqs, obj, kind=kind, **self._interp_kwargs)
+
+    def _check_format(self):
+        if self._obj is None:
```
steven-murray: Instead of supporting different kinds of objects, why not just make a base class and subclasses? That's what they exist for! Then the user can easily define `my_bandpass = Bandpass(...)`.

hera_sim/interpolators.py
```diff
@@ -131,3 +157,106 @@ def _check_pol(self, pol):
...
+            assert self._obj in self._data.keys() and 'freqs' in self._data.keys(), \
+                    "You've chosen to use an interp1d object for modeling the " \
+                    "{}. Please ensure that the `.npz` archive has the following " \
+                    "keys: 'freqs', '{}'".format(self._obj, self._obj)
+        else:
+            # we can relax this a bit and allow for users to also pass a npz
+            # with the same keys as in the above case, but it seems silly to
+            # use a polynomial interpolator instead of a spline interpolator in
+            # this case
+            assert self.datafile[-4:]=='.npy', \
```
steven-murray: there's a function in `os.path` for getting the file extension

hera_sim/io.py
```diff
@@ -4,14 +4,15 @@
 """
 import numpy as np
 from pyuvsim.simsetup import initialize_uvdata_from_keywords
+from .defaults import _defaults
 
 HERA_LAT_LON_ALT = (
```
steven-murray: This should be part of the defaults!

hera_sim/io.py
```diff
@@ -4,14 +4,15 @@
...
     np.degrees(-0.53619179912885),
     np.degrees(0.3739944696510935),
     1073.0000000074506
 )
 
-
+@_defaults
```
steven-murray: This is really cool, however I like the idea of making these `model`s. This is more semantic.

steven-murray: We could do this with classes also, but I kind of like the deco syntax -- and each model really does only have one function call.

hera_sim/noise.py
```diff
@@ -7,26 +7,28 @@
...
-HERA_Tsky_yy = np.concatenate([HERA_Tsky_yy[-10:], HERA_Tsky_yy, HERA_Tsky_yy[:10]])
+from .interpolators import Tsky, _read_npy
+from .defaults import _defaults
+
+HERA_TSKY_VS_LST_NPZ = os.path.join(DATA_PATH, 'HERA_Tsky_Reformatted.npz')
+
 HERA_Tsky_mdl = {}
-HERA_Tsky_mdl['xx'] = RectBivariateSpline(lsts, fqs, HERA_Tsky_xx, kx=4, ky=4)
-HERA_Tsky_mdl['yy'] = RectBivariateSpline(lsts, fqs, HERA_Tsky_yy, kx=4, ky=4)
+HERA_Tsky_mdl['xx'] = Tsky(HERA_TSKY_VS_LST_NPZ, pol='xx')
```
steven-murray: I want these kinds of lines removed completely from the code: the code itself shouldn't care about HERA-specific stuff. That should all live in the config module.

hera_sim/noise.py
steven-murray: Isn't that what your decorator is supposed to take care of?

hera_sim/noise.py
```diff
@@ -7,26 +7,28 @@
...
 
-HERA_BEAM_POLY = np.array([8.07774113e+08, -1.02194430e+09,
-                           5.59397878e+08, -1.72970713e+08, 3.30317669e+07, -3.98798031e+06,
-                           2.97189690e+05, -1.24980700e+04, 2.27220000e+02])  # See HERA Memo #27
+# turns out that this will just return the H1C beam polynomial. See
+# HERA Memo #27 for info on how the polyfit was obtained.
+HERA_BEAM_POLY = _get_hera_bm_poly()
 
-def bm_poly_to_omega_p(fqs, bm_poly=HERA_BEAM_POLY):
+def bm_poly_to_omega_p(fqs, bm_poly=_get_hera_bm_poly):
```
steven-murray: I think the typical solution to this is to set `bm_poly=None` in the sig, and then evaluate it in the function body.

hera_sim/noise.py
```diff
@@ -173,6 +177,9 @@ def thermal_noise(fqs, lsts, Tsky_mdl=None, Trx=0, omega_p=None, inttime=10.7, *
     """
     if omega_p is None:
```
steven-murray: why not just change this call to _get_hera_beam_poly()(fqs)?

hera_sim/rfi.py
```diff
@@ -69,36 +86,8 @@ def gen_rfi(self, fqs, lsts, rfi=None):
...
-    (0.1911, 1.0, 1000 * 100.0, 30.0, 1000),
-    (0.1972, 1.0, 1000 * 100.0, 30.0, 1000),
-    # DC Offset from ADCs
-    # (.2000, 1., 100., 0., 10000),
-]
-
-
 # XXX reverse lsts and fqs?
-def rfi_stations(fqs, lsts, stations=HERA_RFI_STATIONS, rfi=None):
+def rfi_stations(fqs, lsts, stations=_get_hera_stations(), rfi=None):
```
steven-murray: Again, use None, and call the func inside the func body. Though, I would have thought that applying `@_defaults` should have dealt with all of this.

hera_sim/sigchain.py
```diff
@@ -39,7 +38,7 @@ def gen_bandpass(fqs, ants, gain_spread=0.1):
     See Also:
         :meth:`~gen_gains`: uses this function to generate full gains.
     """
-    bp_base = np.polyval(HERA_NRAO_BANDPASS, fqs)
+    bp_base = np.polyval(_get_hera_bandpass(), fqs)
```
steven-murray: this should be a parameter to the function, and again, set its default to `None`.

I really don't think you should even need to define a `_get_hera_bandpass` function: the `@_defaults` should have the relevant logic to apply this automatically, if it finds such an argument in the signature.

hera_sim/tests/test_defaults.py
```diff
@@ -0,0 +1,42 @@
+from os.path import join
+import numpy as np
+from hera_sim.defaults import defaults
+from hera_sim.config import CONFIG_PATH
+from hera_sim.sigchain import gen_bandpass
+from hera_sim.noise import bm_poly_to_omega_p
+
+if defaults._version_is_compatible:
+    def test_config_swap():
+        defaults.set_defaults('h1c')
```
steven-murray: you can probably make the method name just `set`, so the call will be `defaults.set('h1c')`. I think that reads more nicely.

hera_sim/tests/test_defaults.py
```diff
@@ -0,0 +1,42 @@
...
+        config1 = defaults._config
+        defaults.set_defaults('h2c')
+        assert config1 != defaults._config
+
+    def test_direct_config_path():
+        config = join(CONFIG_PATH, 'HERA_H2C_CONFIG.yaml')
+        defaults.set_defaults(config)
```
steven-murray: since you know some of the values in this config, perhaps assert that a few of them are what you expect.

hera_sim/tests/test_defaults.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def test_beam_poly_changes():
+        defaults.set_defaults('h1c')
+        defaults.activate_defaults()
```
steven-murray: I would argue that merely calling `defaults.set('h1c')` here should be enough to also `activate` them. If the user is explicitly setting them, they expect them to be active. There should still be the option to deactivate and activate manually, though.

hera_sim/noise.py
r-pascua: It is, but it looks like the decorator doesn't work for situations like this. However, I did figure out a workaround. So currently an object like `HERA_BEAM_POLY` gets permanently set to whatever the initial default setting is (currently H1C). I think that's okay, though, because we shouldn't be storing data in the code anyway; I see their existence from here onward as more of a legacy feature, kept primarily for historical purposes.

hera_sim/config/HERA_H1C_CONFIG.yaml
```diff
@@ -0,0 +1,66 @@
...
+    gen_bandpass:
+        gain_spread: 0.1
+    gen_whitenoise_xtalk:
+        amplitude: 3.0
+    gen_cross_coupling_xtalk:
+        amp: 0.0
+        dly: 0.0
+        phs: 0.0
+
+# XXX do we touch anything in utils, vis?
```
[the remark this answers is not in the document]
r-pascua: I agree that 3 is probably the best, but it may be good to get input from other folks on this.

hera_sim/config/HERA_H2C_CONFIG.yaml
```diff
@@ -0,0 +1,67 @@
+foregrounds:
```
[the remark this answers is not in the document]
r-pascua: I like this idea.

I think that this would involve substantial modifications to the `Defaults` class, but this approach agrees more with the concept behind the functionality implemented by the class. I can look into implementing this if we think it's worth pursuing.

hera_sim/defaults.py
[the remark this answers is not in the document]
r-pascua: There already is a `set_defaults` method, which changes the location of the configuration file pointed to. If `Defaults._use_season_defaults` is `True` (which is what happens when `Defaults.activate_defaults()` is called), then these new defaults will be used everywhere.

I'll update the initial comment in this PR to outline the basic functionality of the class and new additions so that it's very clear what the class actually does.

hera_sim/config/HERA_H2C_CONFIG.yaml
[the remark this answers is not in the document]
steven-murray: let's leave it for a while, and keep it in mind, since it will take a bit of work. I think most of the other ideas don't really depend on this, so it should be okay to leave it for a while to stew.

hera_sim/config/HERA_H1C_CONFIG.yaml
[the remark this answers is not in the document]
steven-murray: I think if you and I agree, let's go with that. I guess the kinds of parameters that this definition excludes are those that are reasonable to have defaults, but don't really depend on a kind of instrument or analysis. These should just have defaults in the func sig, not in the YAML.

For example, defaults of 0.0 for parameters are signals that those parameters don't really have telescope-dependent defaults, but rather defaults that just "make sense".

hera_sim/defaults.py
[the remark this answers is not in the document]
steven-murray: Yeah, that became more clear as I read through the code, but it would be good to front-load the example usage :-).

I still stand behind the idea that _explicitly_ calling `set_defaults` (or `set()`, as I would have it) should by default also set `activate_defaults` (or `activate()` as I would have it).

Also, just thinking, is `Defaults` meant to be a singleton? I.e. is `defaults` meant to be the only ever instance of `Defaults`? The user is not supposed to make their own instance, right, just to use the pre-made one? In that case, we should hide `Defaults --> _Defaults`.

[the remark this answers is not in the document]
r-pascua: Yes, `Defaults` is meant to be a singleton. I suppose we could hide it, but the package `__init__` method only loads in the `defaults` instance of `Defaults` from the `defaults` module (sorry if that sounds confusing), so that the `defaults` module is not discoverable and an end-user should not be able to make their own `Defaults` instance.

hera_sim/tests/test_defaults.py
r-pascua: I think that's sensible.

hera_sim/defaults.py
```diff
@@ -0,0 +1,185 @@
...
+
+SEASON_CONFIGS = {'h1c': path.join(CONFIG_PATH, 'HERA_H1C_CONFIG.yaml'),
+                  'h2c': path.join(CONFIG_PATH, 'HERA_H2C_CONFIG.yaml'),
+                  }
+
+class Defaults:
+    """
+    This class handles the retreival of simulation default parameters from
+    YAML files and the ability to switch the default settings while in an
+    interactive environment.
```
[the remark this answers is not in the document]
r-pascua: Will do.

hera_sim/config/HERA_H1C_CONFIG.yaml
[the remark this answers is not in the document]
r-pascua: I'm thinking it may be good to modularize a bit; if we change the implementation to option (3), then it may be good to have the YAML configured something like:
```
instrument:
    beam: !beam
    ...
analysis:
    ...
```
What do you think?

hera_sim/defaults.py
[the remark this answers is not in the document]
steven-murray: I would still hide it with an underscore, as that is the convention.

hera_sim/config/HERA_H1C_CONFIG.yaml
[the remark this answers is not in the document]
steven-murray: yes, agree!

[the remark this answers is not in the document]
r-pascua: Regarding the `@component` and `@model` features, do you think it's a good idea to include these features in this PR? It seems like a somewhat independent feature from what I'm aiming to achieve in this PR, so I think it may be more appropriate to create another branch to implement these features.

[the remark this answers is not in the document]
steven-murray: I feel like it's very much related, but perhaps you're right. We do want to be careful that we don't have too many PRs implementing _breaking_ changes though. If you think that implementing those will not break anything further, then definitely separate it.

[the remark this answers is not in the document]
r-pascua: I suspect that I may be able to implement the `@model` and `@component` decorators without breaking anything, but it's hard to say without having actually begun development.

On another note, unless I'm missing some small details, the current state of this branch does not break any of the pre-existing API. I have been trying my best to be careful while developing and ensure that all of the code in `hera_sim` can be used in exactly the same way as it is currently used.

hera_sim/interpolators.py
r-pascua: Ah, no, the way I was envisioning it was that this would add a keyword 'object' whose value specified whether the interpolation object represents a beam or a bandpass. Although I now realize this is not a good approach.

hera_sim/defaults.py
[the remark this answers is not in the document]
r-pascua: I've just added an `Examples` section. Please look it over and let me know if there are any additional examples you think I should add.

hera_sim/interpolators.py
r-pascua: Is it a bad assumption to make if the error message states that the file is assumed to exist in the `hera_sim.data` directory (when `datafile` is not an absolute path)?

hera_sim/defaults.py
r-pascua: Resolved by making custom YAML tags.

r-pascua:
Notes from discussion on 2019-08-22:
- As long as there is no API breaking, tag this PR as the last minor version before next major version.
- Add deprecation warning informing users that HERA-specific variables will be moved in the next major release.
- Improve implementation of `interpolators.freq_interp1d`?
- Postpone implementation of #25 to next major release. Some additional features desired for the next major release are:
-- Make the `Simulator` class be the _de jure_ and _de facto_ primary interface for `hera_sim`.
-- Modify the way defaults and simulation configurations are handled to be more semantic (i.e. configurations have an `analysis` component and an `instrument` component).
-- Make class-like objects for storing HERA-specific variables and configurations. Generalize some helper functions (e.g. replace `noise.bm_poly_to_omega_p` by some `get_omega_p`) and use these class-like objects to define the particular way these helper functions are implemented.
-- Address #23, #24, and #55; fine-tune updates implemented in #46.

@steven-murray @nkern, am I missing anything?

hera_sim/config/HERA_H1C_CONFIG.yaml
```diff
@@ -1,12 +1,12 @@
...
-        Tsky_mdl: 
+        Tsky_mdl: !Tsky 
             datafile: HERA_Tsky_Reformatted.npz
             interp_kwargs: {}
-        omega_p:
+        omega_p: !freq_interp1d
             datafile: HERA_H1C_BEAM_POLY.npy
             interp_kwargs:
-                object: beam
+                obj: beam
```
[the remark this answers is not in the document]
r-pascua: `bandpass` and `beam` are currently the only acceptable values. I'm thinking of removing this behavior entirely, though, as it's only put in there as a convention for archived array names in `.npz` files used to generate 1d spline interpolators.

hera_sim/config/HERA_H1C_CONFIG.yaml
```diff
@@ -23,13 +23,13 @@ noise:
     sky_noise_jy:
         inttime: 10.7
     thermal_noise:
-        Tsky_mdl:
+        Tsky_mdl: !Tsky
```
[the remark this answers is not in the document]
r-pascua: I agree that the semantic version is better, especially for things like this. Getting that working will require substantial effort though, and I'm viewing the state of `hera_sim` after this PR as something of an outline of all the features we would like to have, even if they don't exactly work how they're supposed to (the `Simulator.add_foregrounds` method is one example of the idea being there but not working correctly) or aren't implemented the best way they can be (the `_Defaults` class as an example here). So I'm thinking it's good to be aware of this, but to save the real effort for making the `v1.0.0` release, since that seems more along the lines of "let's actually implement everything the right way and the best way we can think of".

hera_sim/interpolators.py
```diff
@@ -168,22 +164,23 @@ class is intended to be used primarily as a helper function for various
...
+    Initialize an interpolation object from a given reference file and choice of 
     interpolator.
 
     Args:
-        ref_file (str):
-            Absolute path to the reference file to be used for making the
-            BeamSize interpolator. This must be either a .npy or .npz file,
-            depending on which interpolation method is chosen.
+        datafile (str):
+            Hook to the reference file to be used for making the 1d-interpolator
```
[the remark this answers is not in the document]
r-pascua: agreed

hera_sim/interpolators.py
```diff
@@ -25,6 +25,13 @@ def _check_path(datafile):
 def _read_npy(npy):
     return np.load(_check_path(npy))
 
+# TODO: Update module to have a base `interpolator` class, from which `Tsky`
```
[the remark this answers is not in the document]
r-pascua: I think so. I'd have to relearn good inheritance practices in Python, so it might delay this release a bit, but it's certainly possible (and I should do the thing anyway).

CHANGELOG.rst
```diff
@@ -1,3 +1,92 @@
...
+  - The following modules have been affected by this API change:
+    - ...**
+
+v0.1.0 [DATE]
+=============
+
+Added
+-----
+
+- :module: interpolators
```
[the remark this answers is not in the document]
r-pascua: gotcha. Do we document details of new things in the change log, or do we just expect people to go and read the documentation to learn about the new things? (That brings up another point--the documentation on the readthedocs is basically nonexistent; should I make an effort to fill the documentation out in this PR, or save that for the `v1.0.0` release?)

CHANGELOG.rst
```diff
@@ -1,3 +1,92 @@
 =========
 Changelog
 =========
+
+
+Unreleased
```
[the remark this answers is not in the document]
r-pascua: OK, but then I'm kind of confused about what the "Unreleased" section is actually for. Do we have a document somewhere within HERA/RASG that details how we version things and how we've agreed to document our code? Like, does it really make sense to say something is unreleased if it's been merged into master? I would think that the "unreleased" section would tell users what they can expect in the future (or at least, that's what I would like to be detailed in the "unreleased" section), but again I'm still pretty new to this whole collaborative development process.

r-pascua:
> This looks really great! I made a few small comments which I think should be easy to resolve. I think my main concerns are getting the right interface for the interpolators, and also for the `_Defaults` class.
> 
> Perhaps these can be fixed better in the 1.0.0 PR though, if you like.

Great! Some of these are rather straightforward fixes, whereas some might take a little bit of time for me to figure out. I think what I'll do is make a copy of this branch to start work for the 1.0.0 release. Anything I get figured out before some of my responses to your comments are addressed, I can just copy over to this branch.

PR #59, opened 2019-08-29
Visibility simulators gpu

Jackmastr:


hera_sim/visibilities/healvis_wrapper.py
```diff
@@ -49,7 +69,7 @@ def sky_model(self):
 
         # convert from Jy/sr to K
         intensity = 10**-26 * self.sky_intensity.T
-        intensity *= (cnst.c.to("m/s").value/self.sky_freqs)**2 / (2 * cnst.k_B.value)
+        intensity *= old_div((old_div(cnst.c.to("m/s").value,self.sky_freqs))**2, (2 * cnst.k_B.value))
```
r-pascua: What's the motivation for using old division here?

hera_sim/visibilities/prisim_wrapper.py
```diff
@@ -186,7 +211,7 @@ def run_region_of_interest(self):
         for i, (ind, time) in enumerate(zip(indices, self.times)):
             if len(ind) == 0:
                 roi.append_settings(
-                    None, self.uvdata.freq_array[0] / 1e9,
+                    None, old_div(self.uvdata.freq_array[0], 1e9),
```
r-pascua: Again, I don't understand why old division is being used here.

hera_sim/io.py
```diff
@@ -82,6 +82,11 @@ def empty_uvdata(nfreq, ntimes, ants, antpairs=None, pols=['xx', ],
 
     # Generate the antpairs if they are not given explicitly.
     antpairs, ant1, ant2 = _get_antpairs(ants, antpairs)
+    
+    def unique(arr):
```
r-pascua: Is this used anywhere?

hera_sim/tests/test_vis.py
```diff
@@ -5,7 +5,11 @@
 
 from hera_sim import visibilities as vis, io
 
-SIMULATORS = (vis.HealVis, vis.VisCPU)
+try:
+    SIMULATORS = (vis.HealVis, vis.VisCPU)
+except AttributeError: # If healvis is not imported
```
r-pascua: I don't think this is the type of error we would want to let pass silently, especially if we're testing the functionality of these simulators. At the very least, I would recommend adding a warning if the `except` clause is executed. Something along the lines of this:
```
except AttributeError:
    warnings.warn("healvis failed to import in the package constructor.")
    ...
```

hera_sim/tests/test_vis.py
```diff
@@ -5,7 +5,11 @@
 
 from hera_sim import visibilities as vis, io
```
r-pascua: I'm not sure if I'm missing something, but why would you give a single module two aliases?

hera_sim/tests/test_vis.py
```diff
@@ -24,6 +28,59 @@ def uvdata():
...
+        point_source_pos=point_source_pos,
+        nside=2 ** 4
+    ).simulate()
+
+    assert viscpu1.shape == viscpu2.shape
+    assert not np.allclose(viscpu1, viscpu2, atol=0.1)
+
+@pytest.fixture
+def uvdata2():
+    return io.empty_uvdata(
```
r-pascua: Maybe not the best place to put this comment, but in the future this should be changed, since `io.empty_uvdata` is quite different in `master`.

hera_sim/tests/test_vis.py
```diff
@@ -40,10 +97,10 @@ def test_shapes(uvdata, simulator):
     v = simulator(
         uvdata=uvdata,
         sky_freqs=np.unique(uvdata.freq_array),
-        sky_intensity=I_sky
+        sky_intensity=I_sky,
     )
 
-    assert v.simulate().shape == (NTIMES, 1, NFREQ, 1)
+    assert v.simulate().shape == (NTIMES*len(uvdata.get_antpairs()), 1, NFREQ, 1) ############################ADDED NUMBASELINES
```
r-pascua: More of a comment on coding style: if you're going to place a comment immediately after a piece of code, a single hash followed by the comment should be just fine:
```
line of code # brief description of line
```
Additionally, `git` keeps track of changes between commits, so comments noting changes like this are not necessary (in fact, we would want to *remove* these comments before merging with `master`, since they'd mainly act as a source of confusion/distraction for other users looking through the source code).

hera_sim/tests/test_vis.py
```diff
@@ -92,26 +149,25 @@ def test_autocorr_flat_beam(uvdata, simulator):
...
-        uvdata=uvdata,
-        sky_freqs=np.unique(uvdata.freq_array),
-        sky_intensity=I_sky,
-    ).simulate()
-
-    # Ensure that increasing sky resolution smooths out
-    # any 'wiggles' in the auto-correlations of a flat sky.
-    assert np.std(np.abs(v)) >= np.std(np.abs(v2))
-
+### NEED TO CONSIDER WHETHER THIS TEST IS APPROPRIATE###
```
r-pascua: If this tests behavior that we expect the simulator to have, then this is certainly a test we should have in place. If we can analytically place some constraints on the results of this test, then we should include those tests as well.

hera_sim/tests/test_vis.py
```diff
@@ -92,26 +149,25 @@ def test_autocorr_flat_beam(uvdata, simulator):
...
+#     I_sky = create_uniform_sky(nbase=6)
+#     v2 = vis.VisCPU(
+#         uvdata=uvdata,
+#         sky_freqs=np.unique(uvdata.freq_array),
+#         sky_intensity=I_sky,
+#     ).simulate()
+
+#     # Ensure that increasing sky resolution smooths out
+#     # any 'wiggles' in the auto-correlations of a flat sky.
+#     assert np.std(np.abs(v)) >= np.std(np.abs(v2))
```
r-pascua: If this test isn't passing, then perhaps we need to rethink how we would test for "amount of wiggles".

hera_sim/visibilities/__init__.py
```diff
@@ -13,12 +13,19 @@
 # Registered Simulators
 from .vis_cpu import VisCPU
 
+
 try:
     from .healvis_wrapper import HealVis
-except ImportError:
+except (ImportError, NameError) as e: # NameError arises bc healvis is not Python 3 compliant
```
r-pascua: Again, I don't think these are the types of errors we want to let pass silently. At the very least, we should be issuing a warning to the user that the simulator failed to be imported. On that note, I think it is worth investigating what is causing the import error. Finally, the note that "healvis is not Py3 compliant" is concerning; since Py2 is being deprecated at the end of the year, we are moving towards dropping Py2 support everywhere and making things compatible with only Py3. So I'm a little suspicious about this claim--it would be helpful to see the actual error message that pops up with the `NameError` so that we can debug.

.travis.yml
```diff
@@ -36,8 +36,9 @@ install:
   - conda create -q -n test-env python=$TRAVIS_PYTHON_VERSION
   - conda env update -q -n test-env -f environment.yml
   - conda activate test-env
+  - pip install -e .
```
r-pascua: Don't think we need the `-e` flag, but I suppose it doesn't really matter.

hera_sim/tests/test_noise.py
```diff
@@ -33,6 +33,8 @@ def test_resample_Tsky(self):
         # import uvtools, pylab as plt
         # uvtools.plot.waterfall(tsky); plt.show()
 
+### Does not pass, needs to be commented for the pull request
+'''
```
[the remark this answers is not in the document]
Jackmastr: `test_noise` failed before I created the `visibility-simulators-GPU` branch (and _still_ fails on the `visibility-simulators` branch).

hera_sim/visibilities/healvis_wrapper.py
Jackmastr: I used the `futurize` command to automatically convert python 2 files to python 3 -- it must have automatically replaced every instance of division with `old_div`

Jackmastr: I can replace them back though, makes the code less clear

hera_sim/io.py
Jackmastr: no -- I've removed it

hera_sim/visibilities/healvis_wrapper.py
r-pascua: Ah, gotcha--the `futurize` command thought that divisions were intended to be the old division. I think that we should only have `old_div` be used if there is a case where we absolutely want to be using it. As far as I know, there aren't any instances where we would prefer to use it, so it shouldn't be there.

hera_sim/visibilities/__init__.py
Jackmastr: OK, I can work to make it a warning. 

This is the error in question (python 3 renamed `unicode` type to just `str`):
https://stackoverflow.com/questions/19877306/nameerror-global-name-unicode-is-not-defined-in-python-3

This `healvis` commit will fix the issue when the pull request for the `major-revisions` branch is approved:
https://github.com/RadioAstronomySoftwareGroup/healvis/pull/28/commits/4a2c2e90682b9831f66522e2c477d191a2ae3b8b

hera_sim/tests/test_noise.py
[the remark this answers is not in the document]
r-pascua: Ah, so it's an old issue. On your local machine, can you navigate to the top-level `hera_sim` directory and run the following sequence of commands?
```
git checkout master
git pull
git checkout visibility-simulators-GPU
git checkout master hera_sim/noise.py hera_sim/tests/test_noise.py
git commit
pytest hera_sim/tests/test_noise.py
```
I believe this should update the `noise` module and the tests that go with it, so that the test shouldn't fail anymore. If it worked, then the tests run with the `pytest` command should pass.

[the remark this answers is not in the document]
Jackmastr: OK, so did that, realized I needed to also pull the `interpolators.py`,  `defaults.py` files, the `config` folder, and many of the items in the `data` folder from the master branch (`visibility-simulators` is quite behind it seems...). 

image

This is the error I am getting now.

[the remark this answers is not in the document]
r-pascua: Ah, crap, that's right--I made a bunch of additions and modifications to `hera_sim` over the past few months. I'd say just do 
```
git checkout d241bdd hera_sim/noise.py hera_sim/tests/test_noise.py
```
(from the top-level `hera_sim` directory) to revert the changes, then run the tests again and make sure everything passes. Since the `visibility-simulators` branch is so far behind `master`, I think it's OK to leave this test commented out for now. Once we get this branch merged in with `visibility-simulators`, we can merge the differences in `master` and deal with it then.

hera_sim/visibilities/__init__.py
r-pascua: Gotcha. In that case, I'd suggest minimal changes:
```
try:
    from .healvis_wrapper import HealVis
except ImportError, NameError:
    warnings.warn("healvis failed to import.")
```
or something along the lines of that. You'll need to `import warnings` as well (but do this with the rest of the imports).

hera_sim/tests/test_noise.py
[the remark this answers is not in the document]
Jackmastr: Done.

hera_sim/visibilities/__init__.py
Jackmastr: Done.

hera_sim/tests/test_vis.py
Jackmastr: Done.

hera_sim/tests/test_vis.py
Jackmastr: Should I include this if I've added the warning in the __init__ file that it did not import?

hera_sim/tests/test_vis.py
Jackmastr: That's funny, because it's exactly how I thought the syntax there worked, but it actually is equivalent to:
```python
from hera_sim import visibilities as vis
from hera_sim import io
```

hera_sim/tests/test_vis.py
r-pascua: I suppose it's not necessary, but it couldn't hurt to throw it in there. That way when `pytest` gives it report, we'll see that there was a warning specifically in this test.

hera_sim/tests/test_vis.py
r-pascua: I see; that actually makes sense. That said, I think that the equivalent statements (what you posted in a code block) is preferable, even if both implementations do the same thing. The two import statements are more readable than a single import statement.

docs/conf.py
```diff
@@ -12,11 +12,12 @@
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 #
+from __future__ import print_function
 import os
 import sys
 
 sys.path.insert(0, os.path.abspath('../'))
-print("ABSPATH: ", os.path.abspath('../'))
+print(("ABSPATH: ", os.path.abspath('../')))
```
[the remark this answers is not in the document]
Jackmastr: `futurize` used when it shouldn't have been. Fixed now.

hera_sim/tests/test_vis.py
Jackmastr: Done.

.travis.yml
Jackmastr: OK, just dropped it to see if it would change anything.

hera_sim/tests/test_vis.py
Jackmastr: Well that's going to be fun... Thanks for letting me know.

hera_sim/tests/test_vis.py
Jackmastr: Ok, made the change.

.travis.yml
Jackmastr: Alright the `-e` flag is needed for Travis to pass. Not sure why.

hera_sim/visibilities/healvis_wrapper.py
Jackmastr: Removed all cases of old_div, I believe.

steven-murray: All code in the repo has been written with py3 in mind (though it currently explicitly supports py2). So there shouldn't be _any_ `old_div`. Just make sure you do a "Search in path" on the repo to find all occurences.

Can I ask why you ran futurize?

hera_sim/tests/test_vis.py
steven-murray: Yes, this test should definitely pass... increasing sky resolution is not going to make the std go _down_.

r-pascua: I guess the important question now is if we want to fix this test in this PR, or fix it in the PR that merges `visibility-simulators` into `master`

hera_sim/visibilities/__init__.py
steven-murray: Can we please instead make it:

```
try:
    from .healvis_wrapper import HealVis
except ImportError: # healvis not installed, that's fine
    HealVis = None
except NameError:
    warnings.warn("healvis failed to import, as it currently is only Python-2 compatible")
```

hera_sim/visibilities/conversions.py
```diff
@@ -53,8 +53,8 @@ def uvbeam_to_lm(uvbeam, freqs, n_pix_lm=63, trunc_at_horizon=False, **kwargs):
     Returns:
         ndarray, shape[nfreq, beam_px, beam_px]: the beam map cube.
     """
-
-    l = np.linspace(-1, 1, n_pix_lm, dtype=np.float32)
+    
+    l = np.linspace(-1, 1, n_pix_lm, dtype=np.float64)
```
steven-murray: are you sure about this? Since `np.float32` is not default, I'm sure I had a good reason to set it explicitly to that.

r-pascua:
This is looking much better. There are still some stylistic changes I would like to see made, but I suppose those aren't necessary for getting this branch merged into `visibility-simulators`. At this point, I'd recommend fixing the conflict in `conversions` and then make any changes suggested by @steven-murray .

hera_sim/visibilities/conversions.py
```diff
@@ -64,16 +64,27 @@ def uvbeam_to_lm(uvbeam, freqs, n_pix_lm=63, trunc_at_horizon=False, **kwargs):
 
     az = -np.arctan2(m, l)
     za = np.arcsin(n)
-
+    
+    # This does not work at the moment
+    ####uvbeam.efield_to_power()
```
steven-murray: This will work once https://github.com/RadioAstronomySoftwareGroup/pyuvsim/pull/232 is merged (hopefully today)

hera_sim/visibilities/healvis_wrapper.py
```diff
@@ -28,9 +31,25 @@ def __init__(self, fov=180, nprocesses=1, sky_ref_chan=0, **kwargs):
 
         # A bit of a hack here because healvis uses its own AnalyticBeam,
         # and doesn't check if you are using pyuvsim's one. This should be fixed.
+
```
steven-murray: remove excessive whitespace lines

hera_sim/visibilities/healvis_wrapper.py
```diff
@@ -59,9 +78,11 @@ def sky_model(self):
     @cached_property
     def observatory(self):
         """A healvis :class:`healvis.observatory.Observatory` instance"""
-        return setup_observatory_from_uvdata(
```
steven-murray: I think just returning it is fine -- no need for an extra variable

hera_sim/visibilities/prisim_wrapper.py
```diff
@@ -46,9 +51,28 @@ def interferometer(self):
...
         """
+
+        ###################################################################
+        antpos, ants = self.uvdata.get_ENU_antpos()
+
+        labels=np.asarray(self.uvdata.get_antpairs(), dtype=[("A2", int), ("A1", int)])
+
+        antpairs = self.uvdata.get_antpairs()
+        baseline_vectors = np.array([antpos[ant[0]] - antpos[ant[1]] for ant in antpairs]) ### ASK
+        print("BASELINE VECTORS", baseline_vectors)
```
steven-murray: Remove print statements from production code

hera_sim/visibilities/prisim_wrapper.py
```diff
@@ -46,9 +51,28 @@ def interferometer(self):
...
+        print("LABELS", labels)
+        print("DTYPE", labels.dtype)
+        print("TYPE", type(labels))
+        print("SHAPE", labels.shape)
+        ###################################################################
+
         return intf.InterferometerArray(
-            labels=np.asarray(self.uvdata.get_antpairs(), dtype=[("A2", int), ("A1", int)]),
-            baselines=self.uvdata.get,  # (N,3) array ENU ask bryna
+            #labels=list(np.asarray(self.uvdata.get_antpairs(), dtype=[("A2", int), ("A1", int)])), ##################### CAST AS LIST/TUPLE?
```
steven-murray: can we get these comments cleaned up?

hera_sim/visibilities/prisim_wrapper.py
```diff
@@ -279,11 +303,14 @@ def _simulate(self):
...
+            # UVData has a bandpass_array, but AnalyticBeams do not
+            bandpass = getattr(self.beams[0], "bandpass_array", np.ones_like(self.uvdata.freq_array[0]))
+
             self.interferometer.observe(
                 timeobj=self._astropy_times[j],
                 Tsysinfo={"Tnet": 0},
-                bandpass=self.beams[0].bandpass_array,
-                pointing_center=np.array([90.0, 270.0]),
+                bandpass=bandpass,
+                pointing_center=np.array([90.0, 270.0]), ### ASK
```
steven-murray: what did you want to ask?

hera_sim/visibilities/prisim_wrapper.py
```diff
@@ -299,13 +326,26 @@ def _simulate(self):
                 vmemavail=None
             )
 
+        #interData = intf.InterferometerData(prisim_object=self.interferometer)
+        #return interData.createUVData()
+
+
         # Reshape array into UVData.data_array format
-        vis = self.interferometer.skyvis_freq.conj()
+        vis = self.interferometer.skyvis_freq.conj() ###############333333
```
steven-murray: clean up comments and print statements

hera_sim/visibilities/vis_cpu.py
```diff
@@ -14,7 +16,7 @@
 class VisCPU(VisibilitySimulator):
 
     def __init__(self, bm_pix=100, real_dtype=np.float32,
-                 complex_dtype=np.complex64, **kwargs):
+                 complex_dtype=np.complex64, c=299792458.0, **kwargs):
```
steven-murray: Can't the c constant be hard-coded or loaded from astropy?

hera_sim/visibilities/vis_cpu.py
```diff
@@ -30,19 +32,23 @@ def __init__(self, bm_pix=100, real_dtype=np.float32,
...
         self.freqs = self.uvdata.freq_array[0]
 
     @property
     def lsts(self):
         try:
             return self.__lsts
-        except AttributeError:
-            self.__lsts = np.unique(self.uvdata.lst_array)
+        except AttributeError:      
+            self.__lsts = self.uvdata.lst_array[::self.uvdata.Nbls] ####np.unique(self.uvdata.lst_array)
```
steven-murray: remove comments at end of line

hera_sim/visibilities/vis_cpu.py
```diff
@@ -70,11 +76,16 @@ def get_beam_lm(self):
             method can be modified to only return one matrix for each beam.
 
         """
+        #print ("self.freqs", self.freqs)
```
steven-murray: remove comments

hera_sim/visibilities/vis_cpu.py
```diff
@@ -90,11 +101,14 @@ def get_eq2tops(self):
         coords to topocentric at each LST.
         """
 
-        sid_time = np.unique(self.uvdata.lst_array)
+        sid_time = self.uvdata.lst_array[::self.uvdata.Nbls]
```
steven-murray: can replace this with `self.lsts`

hera_sim/visibilities/vis_cpu.py
```diff
@@ -90,11 +101,14 @@ def get_eq2tops(self):
...
         eq2tops = np.empty((len(sid_time), 3, 3), dtype=self._real_dtype)
-
+        
+        #ra, dec = self.point_source_pos.T
```
steven-murray: remove comment

hera_sim/visibilities/vis_cpu.py
```diff
@@ -90,11 +101,14 @@ def get_eq2tops(self):
...
+        
+        
         for i, st in enumerate(sid_time):
-            eq2tops[i] = conversions.eq2top_m(-st, self.uvdata.telescope_lat_lon_alt[0])
+            eq2tops[i] = conversions.eq2top_m(st, self.uvdata.telescope_lat_lon_alt[0]) ######### WAS -st
```
steven-murray: can you go through the math and show that it should be +st here?

hera_sim/visibilities/vis_cpu.py
```diff
@@ -105,19 +119,21 @@ def _base_simulate(self, crd_eq, I):
         visfull = np.zeros_like(self.uvdata.data_array, dtype=self._complex_dtype)
 
         for i, freq in enumerate(self.freqs):
+
             vis = vis_cpu(
                 antpos=self.antpos,
-                freq=freq,
+                freq=self.freqs[i],
```
steven-murray: this can just be `freq`

hera_sim/visibilities/vis_cpu.py
```diff
@@ -105,19 +119,21 @@ def _base_simulate(self, crd_eq, I):
...
                 eq2tops=eq2tops,
                 crd_eq=crd_eq,
                 I_sky=I[i],
                 bm_cube=beam_lm[:, i],
                 real_dtype=self._real_dtype,
-                complex_dtype=self._complex_dtype
+                complex_dtype=self._complex_dtype,
+                c=self.c
```
steven-murray: not sure why `c` is being passed... it's a constant

hera_sim/visibilities/vis_cpu.py
```diff
@@ -168,6 +184,23 @@ def vis_cpu(antpos, freq, eq2tops, crd_eq, I_sky, bm_cube, real_dtype=np.float32
     Returns:
         array_like, shape(NTIMES, NANTS, NANTS): visibilities
     """
+    
+    ####################################################################
```
steven-murray: rm comments

hera_sim/visibilities/vis_cpu.py
```diff
@@ -200,6 +234,9 @@ def vis_cpu(antpos, freq, eq2tops, crd_eq, I_sky, bm_cube, real_dtype=np.float32
     # Loop over time samples
     for t, eq2top in enumerate(eq2tops.astype(real_dtype)):
         tx, ty, tz = crd_top = np.dot(eq2top, crd_eq)
+        
+        #print "VISCPU CRD_TOP", crd_top
```
steven-murray: rm comments

hera_sim/visibilities/vis_cpu.py
```diff
@@ -209,7 +246,13 @@ def vis_cpu(antpos, freq, eq2tops, crd_eq, I_sky, bm_cube, real_dtype=np.float32
 
         # Calculate delays
         np.dot(antpos, crd_top, out=tau)
-        np.exp((1.0j * ang_freq) * tau, out=v)
+        #TAU = (b * s) / c
+        tau = tau/c
```
steven-murray: use `tau /= c`

hera_sim/visibilities/vis_cpu.py
```diff
@@ -209,7 +246,13 @@ def vis_cpu(antpos, freq, eq2tops, crd_eq, I_sky, bm_cube, real_dtype=np.float32
...
+        
+        
+        np.exp(1.0j * (ang_freq * tau), out=v)
+        
+        #print "VISCPU ARGUMENT OF EXPONENTIAL", ang_freq * tau[1][0]
```
steven-murray: rm comment

hera_sim/visibilities/vis_cpu.py
```diff
@@ -218,11 +261,14 @@ def vis_cpu(antpos, freq, eq2tops, crd_eq, I_sky, bm_cube, real_dtype=np.float32
         for i in range(len(antpos)):
             np.dot(v[i: i + 1].conj(), v[i:].T, out=vis[t, i: i + 1, i:])
 
+        #print("A_s at time t=", t, "is", A_s)
```
steven-murray: rm comment

hera_sim/visibilities/vis_cpu.py
Jackmastr:
```python
def eq2top_m(ha, dec):
    """
    Return the 3x3 matrix converting equatorial coordinates to topocentric
    at the given hour angle (ha) and declination (dec).
```
https://en.wikipedia.org/wiki/Hour_angle#Relation_with_the_right_ascension
image

This is how I understand it, correct me where I'm wrong:
(1) The hour angle is passed as the first argument. (2) The conversion on Wikipedia is hour angle = lst - (right ascension of object) (3) the variable `st` is the lst so it is not negated?

hera_sim/visibilities/conversions.py
Jackmastr: Not sure, so I switched it back to `np.float32`.

hera_sim/visibilities/healvis_wrapper.py
Jackmastr: Done.

hera_sim/visibilities/healvis_wrapper.py
Jackmastr: Done.

hera_sim/visibilities/prisim_wrapper.py
Jackmastr: Done.

hera_sim/visibilities/prisim_wrapper.py
Jackmastr: OK cleaned up those.

hera_sim/visibilities/prisim_wrapper.py
Jackmastr: There was an issue regarding the shape, but we pushed off discussing the `PRISim` wrapper implementation until later anyways.

hera_sim/visibilities/prisim_wrapper.py
Jackmastr: Done.

hera_sim/visibilities/vis_cpu.py
Jackmastr: Done.

hera_sim/visibilities/vis_cpu.py
Jackmastr: Done.

hera_sim/visibilities/vis_cpu.py
Jackmastr: Done.

hera_sim/visibilities/vis_cpu.py
Jackmastr: Done.

hera_sim/visibilities/vis_cpu.py
Jackmastr: Done.

hera_sim/visibilities/vis_cpu.py
Jackmastr: Done.

hera_sim/visibilities/vis_cpu.py
Jackmastr: Done.

hera_sim/visibilities/vis_cpu.py
Jackmastr: Done.

hera_sim/visibilities/vis_cpu.py
Jackmastr: Done.

hera_sim/visibilities/vis_cpu.py
Jackmastr: Done.

hera_sim/visibilities/vis_cpu.py
Jackmastr: Although that would beg the question of why we wouldn't be taking into account the right ascension of the object? I do think I am misunderstanding something here.

hera_sim/visibilities/vis_cpu.py
Jackmastr: Removed it. Was only there temporarily to check that `c` being calculated more or less accurately might have caused some of the phase differences between `vis_cpu` and `healvis` (which was not the case).

hera_sim/visibilities/vis_cpu.py
Jackmastr: Importing it from astropy now.

hera_sim/visibilities/conversions.py
Jackmastr: Awesome! Yes I tried and it does work now.

Jackmastr:
2 telescopes, 1 point source at zenith:
image

3 telescopes, 1 point source at zenith:
image


1 telescope, 1 point source at zenith, "airy" beam of diameter=2:
image

hera_sim/visibilities/vis_cpu.py
Jackmastr: Done.

hera_sim/visibilities/prisim_wrapper.py
steven-murray: Yeah, cool. Actually, I think once we merge this, I'll remove the PRISIM wrapper from the branch into its own branch for later.

hera_sim/visibilities/vis_cpu.py
steven-murray: Yeah, I'm also confused, and I can't remember my thinking. Can you put a comment something like 

```
# TODO: ensure that LST is converted to HA correctly (https://en.wikipedia.org/wiki/Hour_angle#Relation_with_the_right_ascension)
```
above the call? 

We'll then deal with it once merged.

environment.yml
```diff
@@ -24,6 +24,6 @@ dependencies:
   - future #only for python 2
   - pip:
     - cached-property
-    - pyuvsim
+    - git+https://github.com/RadioAstronomySoftwareGroup/pyuvsim.git
```
steven-murray: We'll need to make this consistent with `setup.py` as well

PR #60, opened 2019-08-29
Add config package to install

steven-murray:
The config package was not being correctly installed. This didn't fail our tests, because the tests were being run from the repo rather than the installed package.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,4 +1,4 @@
-from setuptools import setup
+from setuptools import setup, find_packages
 import glob
 import os
 import sys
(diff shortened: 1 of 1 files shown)


steven-murray:
Requested @bhazelton since this is currently a block to her own PR.

bhazelton:
Looks reasonable, but I'm not sure I know that it'll work. I assume you've tested offline?

steven-murray:
Yes, it works as intended offline (failing test here is IERS again...)

PR #63, opened 2019-09-19
Run simulation from command line

r-pascua:
Basic framework for running a simulation from the command line; supports BDA in principle, but needs testing.


scripts/simulation.py
```diff
@@ -0,0 +1,96 @@
...
+instrument_params = {"antennas" : antennas}
+for parameter in ("freq", "time", ):
+    for key, value in yaml_contents[parameter].items():
+        instrument_params[key] = value
+
+sim = hera_sim.Simulator(**instrument_params)
+
+# extract simulation parameters
+sim_params = {}
+for component in ("systematics", "analysis", ):
```
r-pascua: The tuple should be reordered so that systematics are simulated last (or we can revert back to having `run_sim` apply effects in a fixed order).

scripts/simulation.py
```diff
@@ -0,0 +1,96 @@
...
+filing_params = yaml_contents["filing"]
+outfile = os.path.join(filing_params["outdir"], filing_params["outfile_name"])
+save_path = "{}.{}".format(outfile, filing_params["output_format"])
+clobber = filing_params["clobber"]
+
+# XXX add a check to make sure the save type is OK?
+
+# determine whether to use season defaults
+defaults = yaml_contents.get("defaults", {})
+apply_defaults = defaults.pop("apply_defaults", False)
```
r-pascua: Can forgo having `apply_defaults` as an entry in the `defaults` dictionary; if `defaults` is not empty, then assume that the user wants defaults set to whatever is specified in the configuration

hera_sim/__yaml_constructors.py
```diff
@@ -21,4 +25,25 @@ def predicate(obj):
...
 interps = dict(inspect.getmembers(interpolators, predicate))
 for tag, interp in interps.items():
-    make_constructor("!%s"%tag, interp)
+    make_interp_constructor("!%s"%tag, interp)
+
+def astropy_unit_constructor(loader, node):
+    params = loader.construct_mapping(node, deep=True)
+    value = params.get("value", None)
+    units = params.get("units", None)
+    # do we want to add handling of values of the form x*np.pi?
```
steven-murray: no

hera_sim/defaults.py
```diff
@@ -51,17 +51,19 @@ class _Defaults:
...
     """Load in a configuration and check its formatting.
 
     Parameters
     ----------
-    config_file : str, optional (default 'h1c') 
-        May either be an absolute path to a configuration YAML or one of
-        the observing season keywords: {}
+    config : str or dict, optional (default 'h1c') 
+        May either be an absolute path to a configuration YAML, one of
+        the observing season keywords ({}), or a dictionary with the
```
steven-murray: not sure what the `({})` is doing here?

hera_sim/defaults.py
```diff
@@ -112,24 +116,43 @@ def deactivate(self):
...
         """Validate the choice of configuration file."""
-        assert isinstance(config_file, str), \
-                "Default configurations are set by passing a path to a " \
-                "configuration file or one of the season keys. The " \
-                "currently supported season configurations are " \
-                "{}.".format(SEASON_CONFIGS.keys())
-
-        if config_file in SEASON_CONFIGS.keys():
-            return SEASON_CONFIGS[config_file]
+        #assert isinstance(config, str), \
```
steven-murray: you can remove the commented-out section

hera_sim/defaults.py
```diff
@@ -112,24 +116,43 @@ def deactivate(self):
...
+        #        "Please ensure that a path to the specified configuration " \
+        #        "file exists. {} could not be found".format(self._config)
+        error_message = "Your configuration should be formatted as a nested " \
+                "dictionary: {module: {model: {params: values}}}"
+        try:
+            module = list(self._config.keys())[0]
+            model = list(self._config[module].keys())[0]
+        except KeyError:
+            raise AssertionError(error_message)
+        # seems a little stupid to use self._config rather than self()
```
steven-murray: so why don't you just use `self(module, model)`?

scripts/simulation.py
steven-murray: Agreed.

scripts/simulation.py
```diff
@@ -0,0 +1,96 @@
...
+if apply_defaults and defaults:
+    if isinstance(defaults["default_config"], str):
+        hera_sim.defaults.set(defaults["default_config"])
+    elif isinstance(defaults["default_config"], dict):
+        hera_sim.defaults.set(**defaults["default_config"])
+    else:
+        raise ValueError("If you wish to override function defaults, then " \
```
steven-murray: or specifying the name of a set of defaults, right?

scripts/simulation.py
```diff
@@ -0,0 +1,96 @@
...
+    sim.data = bda_tools.apply_bda(sim.data, **bda_params)
+
+# save the simulation
+# note that we may want to allow the functionality for the user to choose some
+# kwargs to pass to the write method
+sim.write_data(save_path, 
+               file_type=filing_params["output_format"],
+               **filing_params["kwargs"])
+
+# should be done now
```
steven-murray: lol, you can remove this line

scripts/simulation.py
steven-murray: what are some example entries in "systematics" and "analysis"?

scripts/template_config.yaml
```diff
@@ -0,0 +1,47 @@
+# This document is intended to serve as a template for constructing new
+# configuration YAMLs for use with the `simulation.py` script.
+
+# XXX do we want this to be an exhaustive template (all possible things
```
steven-murray: I think exhaustive is better for a template

scripts/template_config.yaml
```diff
@@ -0,0 +1,47 @@
...
+        value: 0
+        units: 'deg'
+    max_time: !dimensionful
+        value: 0
+        units: 's'
+    corr_int_time: !dimensionful
+        value: !!null
+        units: 's'
+defaults:
+    apply_defaults: False
```
steven-murray: does it make sense to specify defaults and not apply them?

scripts/template_config.yaml
```diff
@@ -0,0 +1,47 @@
...
+    n_freq: 1024
+    channel_width: 122070.3125
+    start_freq: 46920776.3671875
+time:
+    n_times: 10
+    integration_time: 8.59
+    start_time: 2457458.1738949567
+telescope:
+    array_layout: 'antenna_layout.csv'
+    # TODO: allow use with hera_sim.antpos
```
steven-murray: flagging a TODO

scripts/template_config.yaml
```diff
@@ -0,0 +1,47 @@
...
+    # alternatively can specify how to construct it with hera_sim.antpos
+systematics:
+    # TODO: fill out what should go in here
```
steven-murray: flagging a TODO

scripts/template_config.yaml
```diff
@@ -0,0 +1,47 @@
...
+analysis:
+    # TODO: fill out what should go in here
```
steven-murray: flagging a TODO

scripts/simulation.py
r-pascua: This is what has been the bottleneck in finishing development for this branch. I'm not exactly sure what should go in each one, and how it should be handled, so I figured we should have a discussion about this. Additionally, as I was preparing this, I noticed that there is a lot of redundancy between the ways that `defaults` and `run_sim` are handled; I suppose this is OK if we're fine with letting `run_sim` override some of the settings in `defaults`, but I can see that as a potential cause for confusion with end-users.

I'm thinking that "systematics" should contain information about gains and crosstalk at the very least, as well as maybe information about rfi and thermal noise. "analysis" should definitely contain information about the EoR signal and foregrounds. I was thinking it could be a good idea to separate things in a way that "systematics" and "analysis" cover all the bases for simulation specifics, while "defaults" handles things more related to season-specific quantities (like `omega_p`). It's a little tricky, though, since there are some quantities that could easily fall into multiple categories (for example, the bandpass interpolator could live in "systematics" just as well as "defaults", and the sky temperature model live in "analysis" or "defaults").

So the question of where to put things ("defaults", "analysis", or "systematics") is still open. I'd appreciate any suggestions you have on the topic.

hera_sim/defaults.py
r-pascua: I had intended to use string formatting to fill that in, but I realize that I didn't ever actually format the string. I'll just fill in the observing season keywords manually.

hera_sim/defaults.py
r-pascua: Done.

scripts/template_config.yaml
r-pascua: not at all; will update before next push.

scripts/template_config.yaml
r-pascua: apparently already did this and hadn't committed the changes.

hera_sim/defaults.py
r-pascua: done

scripts/simulation.py
r-pascua: updated the error message

scripts/simulation.py
steven-murray: Hmmm. Yeah good point, that there's a lot of crossover between defaults and the other categories. I'm not sure of the best way to handle this. One way could be to assume in this case that `defaults` literally has to be a string path or identifier, and that these parameters really just overwrite those defaults at some places, and add some filing information. 

If most of the parameters are the same between defaults and this config, you can use the same function to interpret them (or at least most of them).

hera_sim/cli.py
```diff
@@ -0,0 +1,103 @@
+"""
+CLI for hera_sim
+"""
+
+import click
+import os
+import yaml
+
+from pyuvsim.simsetup import _parse_layout_csv
+from bda import bda_tools
```
steven-murray: We need to add this to setup.py. Is it really required?

hera_sim/cli.py
```diff
@@ -0,0 +1,103 @@
...
+        outfile = os.path.join(filing_params["outdir"], filing_params["outfile_name"])
+    fmt = filing_params["output_format"]
+    if not outfile.endswith('.%s'%fmt):
+        outfile += ".%s"%fmt
+
+    if os.path.exists(outfile) and not filing_params['clobber']:
+        print("Nothing to do: %s already exists and clobber=False"%outfile)
+        return
+
+    # XXX add a check to make sure the save type is OK?
```
steven-murray: Yeah, let's add this

setup.py
```diff
@@ -59,6 +60,9 @@ def package_files(package_dir, subdirectory):
         'aipy>=3.0',
         'click'
     ],
+    "extras_require" : [
+        'bda @ git+git://github.com/HERA-Team/baseline_dependent_averaging'
```
steven-murray: this should be a dictionary like {"bda": "bda @ git+git://github.com/HERA-Team/baseline_dependent_averaging"}

setup.py
```diff
@@ -59,6 +60,9 @@ def package_files(package_dir, subdirectory):
...
+    ],
     "version": version.version,
     "package_data": {"hera_sim": data_files},
     "zip_safe": False,
```
steven-murray: I think there should be an `entry_points = ...` line in the setup args.

hera_sim/cli.py
```diff
@@ -70,28 +85,35 @@ def run(input, outfile, verbose):
...
+    # the configuration file should only specify any particular parameter once
+    # i.e. the same sky temperature model should be used throughout
     sim_params = {}
-    for key, value in yaml_contents["simulation"].items():
-        sim_params[key] = value
+    for component in yaml_contents["simulation"]["include"]:
+        for content in yaml_contents.values():
+            if component in content.keys():
+                for particular_component, parameters in content[component].items():
+                    sim_params[particular_component] = parameters
```
steven-murray: I'm probably just missing it, but it's not clear to me how this will ensure that each parameter only needs to be specified once? It looks like it will have to be specified for each kind of component that's in `simulation['include']`?

config_examples/template_config.yaml
```diff
@@ -45,24 +43,94 @@ telescope:
         sep: 14.6
         split_core: True
         outriggers: 2
+    # TODO: check how pyuvsim handles beams, see if YAML constructor needs
+    # to be updated/modified.
+    omega_p: !Beam
+        # TODO: add support for specifying directory containing beam file
```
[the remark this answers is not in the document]
r-pascua: yeah. I'm not sure why I was thinking that I would want a feature like that, to be honest

config_examples/template_config.yaml
```diff
@@ -45,24 +43,94 @@ telescope:
...
+            freq_max: 0.214
+            width: 0.008
+            chance: 0.0001
+            strength: 10.0
+            strength_std: 10.0
+    sigchain:
+        gains:
+            gain_spread: 0.1
+            dly_rng: (-20, 20)
+        # XXX: the gen_reflection_gains function is somewhat difficult
```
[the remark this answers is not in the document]
r-pascua: the problem is that reflection gains requires a list for parameters like `amp` that has the same length as the number of antennas, and it defaults to zero if a list is not passed. from the docstring:
```
        amp (list, optional): antenna reflection amplitudes for each antenna. Default is 1.0
        dly (list, optional): antenna reflection delays [nanosec]. Default is 0.0
        phs (lists, optional): antenna reflection phases [radian]. Default is 0.0
```
it says the default amplitude is 1.0, but it's actually 0.0

config_examples/template_config.yaml
```diff
@@ -45,24 +43,94 @@ telescope:
...
+        reflections: # TODO: make a YAML constructor for this?
+            amp: !!null
+            dly: !!null
+            phs: !!null
+    crosstalk:
+        # only one of the two crosstalk methods should be specified
+        whitenoise_xtalk:
+            amplitude: 3.0
+        # cross_coupling_xtalk:
+            # the following parameters are all floats, but it's a little
```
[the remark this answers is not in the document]
r-pascua: the problem here is with how the `add_xtalk` method of the `Simulator` class was written. as it currently stands, every baseline gets the same `kwargs` passed to the crosstalk method, so crosstalk will show up in every visibility waterfall at the same delay, with the same phase and amplitude

config_examples/template_config.yaml
```diff
@@ -45,24 +43,94 @@ telescope:
...
+        interp_kwargs: {}
+        # TODO: add functionality for handling multiple polarizations
+        # this involves updating both the YAML constructor and Simulator
+        #
+    # TODO: investigate how to properly handle EoR here, since this
+    # won't work without modifying some things. 
     eor:
         noiselike_eor:
             eor_amp: 0.00001
             min_delay: !!null
```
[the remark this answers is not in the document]
r-pascua: `None`; trying to specify a value as `None` by writing something like `parameter: None`, then it'll actually be interpreted as `'None'`

hera_sim/cli.py
```diff
@@ -28,7 +32,16 @@ def run(input, outfile, verbose):
...
+
+    # figure out whether or not to do BDA
+    bda_params = yaml_contents.get("bda", {})
+    # make sure bda is installed if the user wants to do BDA
+    if bda_params:
+        raise ImportError("You have defined BDA parameters but do not have "
+                          "bda installed. Please install bda to proceed.")
+
     # extract parameters for saving to disk
+    # XXX: either click options need to be changed, or this does
```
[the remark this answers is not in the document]
r-pascua: ah, nothing. I think I just didn't look at the code closely enough.

hera_sim/cli.py
```diff
@@ -70,28 +85,35 @@ def run(input, outfile, verbose):
...
+    #    for key, value in yaml_contents[component].items():
+    #        config_params[key] = value
+    # default behavior should be such that anything specified in the
+    # configuration parameters will override whatever the current default
+    # settings are, but perhaps warn the user that something is being
+    # specified multiple times if defaults have been set and some of the
+    # defaults conflict with the configuration settings
+    # in the end, we want to handle the configuration stuff with the
+    # functionality offered by defaults
+    hera_sim.defaults.set(**config_params)
```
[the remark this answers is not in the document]
r-pascua: for now, yes, but the intention is that I'll replace the commented out section with code that will fill up `config_params`.

hera_sim/cli.py
r-pascua: I'm pretty sure I was thinking that the `defaults` would take care of this, and that the contents in, say `yaml_contents['sky']['foregrounds']['diffuse_foreground']` would just specify the kwargs that are particular to that method

config_examples/template_config.yaml
```diff
@@ -45,24 +43,94 @@ telescope:
         sep: 14.6
         split_core: True
         outriggers: 2
+    # TODO: check how pyuvsim handles beams, see if YAML constructor needs
+    # to be updated/modified.
+    omega_p: !Beam
```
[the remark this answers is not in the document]
r-pascua: I don't understand what you're getting at? This is in `telescope`, and it seems appropriate to put `omega_p` in the `telescope` section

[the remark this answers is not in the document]
steven-murray: Yeah that would make sense... it may have been the Github formatting, made it look like it was in the `filing` section

config_examples/template_config.yaml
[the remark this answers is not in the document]
steven-murray: Hmmm, yeah array-valued arguments are always going to be tricky. One way I've tried to deal with this before is to have a constructor that basically says "this is going to be a file... read it in". You can even go so far as to have a _list_ of possible ways to read it in (i.e. from `npz`, `npy`, HDF5, ASCII,...), and cycle through them until something works.

config_examples/template_config.yaml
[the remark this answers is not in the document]
steven-murray: Hmmm, OK that sounds like it needs to be fixed then.

config_examples/template_config.yaml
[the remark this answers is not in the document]
steven-murray: right. Can you just omit `min_delay` here (since I assume its default is None)?

hera_sim/cli.py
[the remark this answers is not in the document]
steven-murray: OK

config_examples/template_config.yaml
```diff
@@ -39,29 +36,27 @@ telescope:
...
             interpolator: interp1d
             fill_value: extrapolate
             # if you want to use a polynomial interpolator instead, then
             # interpolator: poly1d
             # kwargs not accepted for this; see numpy.poly1d documentation
-    Trx: 0.0 # receiver temperature for thermal noise
 defaults:
-    # behavior of specifying default config and some other
-    # parameters (e.g. omega_p, Tsky_mdl) currently unknown
+    # XXX: THIS MUST BE A STRING SPECIFYING A DEFAULT CONFIG FILE
```
steven-murray: de-capitalise and remove XXX

hera_sim/cli.py
```diff
@@ -19,15 +21,18 @@
 
 
 @main.command()
-@click.argument('input', type=click.Path(exists=True, dir_okay=False),
-                help="input YAML configuration file")
+@click.argument('input', type=click.Path(exists=True, dir_okay=False))
+#                help="input YAML configuration file")
```
steven-murray: you can remove this

config_examples/template_config.yaml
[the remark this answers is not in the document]
r-pascua: I can, but I wrote the template so that places where function parameters are defined (like here) lists all of the parameters in the signature with their default values. So removing something because its default is the same as what it's set to in the template config doesn't make much sense to me.

config_examples/template_config.yaml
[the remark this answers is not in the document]
r-pascua: It does need to be fixed, but I'd rather fix it in the version 1 release, since there are a few other issues with the `Simulator` class that need to be fixed, and people can just as well use `gen_whitenoise_xtalk` for a quick-and-dirty simulation of crosstalk.

config_examples/template_config.yaml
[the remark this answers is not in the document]
r-pascua: Okay, I suppose we could require that this part be dealt with by specifying the parameters in a file, but I don't think we should just blindly guess at which way the file has been formatted. I think, if we're going to use a parameter file here, that it is best to explicitly state which file format is expected and how the file contents should be arranged.

I've been thinking that it would be a good idea to update the `gen_reflection_gains` function to allow for keywords to be used to make the lists (e.g. 'random' for the phases would choose all of the phases randomly), but I think this is not the place to have that discussion.

Ultimately, I'm not convinced that we absolutely need to solve this issue before merging. The main reason this PR was even made was to provide an interface for quickly generating a mock set of data to be used for benchmarking tests, and I suspect that there isn't a huge demand for including cable reflections in the mock data. So for now, I'd say that anyone who really wants to use this feature should just input the values as a list (no problem doing this with a YAML), and we can work on developing a nicer interface in the future.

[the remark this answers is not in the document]
steven-murray: OK yes I agree -- we can decide on a standard and stick to it. I also agree that it doesn't *need* to be done right now. Let's get this into the wild, and see how it works in practice, and refine from there. In any case, input as a list (in YAML) is something we probably want to support, even if other better methods are also supported.

config_examples/template_config.yaml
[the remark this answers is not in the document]
steven-murray: OK, yup, that makes sense.

config_examples/template_config.yaml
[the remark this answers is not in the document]
steven-murray: OK, I can go with that. I'll make an issue, which can either be addressed independently, or in your v1 PR.

hera_sim/simulate.py
```diff
@@ -231,14 +252,19 @@ def __init__(
...
                     self.data.nsample_array[:] = 1.0
-            elif self.data is not None:
+            else:
                 self.data = data
 
         # Assume the phase type is drift unless otherwise specified.
         if self.data.phase_type == "unknown":
             self.data.set_drift()
 
+        # what does this line do?
```
steven-murray: I have a feeling it might be doing some initialization, but I can't remember for sure. Can you test with it deleted? If nothing bad happens, remove it, if so, comment it!

hera_sim/simulate.py
```diff
@@ -267,10 +293,21 @@ def write_data(self, filename, file_type="uvh5", **kwargs):
                 :class:`pyuvdata.UVData`) which determines which write method to call.
             **kwargs: keyword arguments sent directly to the write method chosen.
         """
+        ret_seeds = kwargs.pop('ret_seeds', False)
```
steven-murray: can we just make these arguments, rather than `kwargs`? I don't see a good reason to leave them as kwargs.

hera_sim/tests/test_simulator.py
```diff
@@ -94,14 +97,21 @@ def test_io():
     sim.add_foregrounds("pntsrc_foreground")
     sim.add_gains()
 
+    # why is this print statement here?
```
steven-murray: just remove it

hera_sim/simulate.py
r-pascua: I would think that it would be better to put them in as named kwargs, rather than something that might be stored in `kwargs` (i.e. including `ret_seeds=False` in the function signature). Is that what you meant?

r-pascua:
I'll go ahead and fix the merge conflicts and address this last batch of comments. I noticed that I have not yet included an example of how to use the CLI; should I do this in the docstring for `run`? I could also make an example notebook that shows how someone can run a simulation from the command line in a way that's consistent with doing the same thing in an interactive environment (thank you `os.system`).

hera_sim/simulate.py
r-pascua: nothing bad happened on removing it, so I removed it for good.

r-pascua:
Removed remaining TODOs and questions from new code written in this PR. Any remaining TODOs or questions/notes (marked by XXX) should have been there prior to this PR, and will be addressed or turned into issues by the version 1 release.

r-pascua:
Ah, one final comment: I just remembered that this PR required the reworking of `defaults` intended for version 1 to be implemented early, and I never got around to updating the docstrings that are now outdated. I also haven't yet updated the changelog, so I'm going to postpone merging this PR until those tasks have been completed. 

Since this does effectively break the API for the `defaults` module, there's also the question of how to appropriately version this. Do we stay true to the versioning convention and release this as 1.0.0, or do we call it 0.2.0 or 0.3.0 (subject to whether #33 is merged before this)? Does anyone except for me even use the features of `defaults` yet?

steven-murray:
Yes, I think there should be an example of how to use this. I'd actually not do it as a notebook, just an RST file. Since you use config files, perhaps do a `cat config.yaml` in the demo, to show its contents. 

I think since `defaults` was merged so recently and I doubt much production code is dependent on it, and since v1 is scheduled soonish, we just pull a fast one this time and release this as 0.2.0 (if you can do it very soon haha).

PR #74, opened 2019-12-14
v1.0.0

### LINKED ISSUE #24: hera specific variables lying around the codebase

### LINKED ISSUE #25: homogenization of model functions

### LINKED ISSUE #55: Simulator.add_foregrounds Polarization Issue
The `Simulator.add_foregrounds` method will not work correctly when trying to simulate diffuse foregrounds for two different polarizations. The current implementation will calculate diffuse foregrounds appropriate for a single polarization and apply the result to all polarizations. For example, if we instantiated a `Simulator` object with `polarization_array=['xx','yy']`, then ran `add_foregrounds('diffuse_foregrounds', Tsky_mdl=hera_sim.noise.HERA_Tsky_mdl['xx'])`, then both the xx- and yy-polarizations would have foregrounds calculated for the xx-polarization.

I can implement a quick-fix for this if we expect this to be the only place we'll add in polarized simulation components; otherwise, we should determine a more general approach to correctly handling multiple polarizations.

### LINKED ISSUE #58: Sphinx documentation
Documentation of various modules and their contents is not automatically pulled from their docstrings.

### LINKED ISSUE #64: problem with add_xtalk
There is a problem with how the  `add_xtalk` method of the `Simulator` class is written. As it currently stands, every baseline gets the same `kwargs` passed to the crosstalk method, so crosstalk will show up in every visibility waterfall at the same delay, with the same phase and amplitude

### LINKED ISSUE #65: Extend Constructor for YAML Tag !antpos
The !antpos constructor currently allows for specification of an array layout using `hera_sim.antpos` functions with parameters defined in the configuration YAML, but it currently does not support specification of an antenna layout csv (or similar file). To simplify the definition of an antenna array in the `cli` module's `run` function, the !antpos tag should be extended to handle array specification from a csv. An alternative is to create a new YAML tag !antcsv (or something similar) that provides this functionality.

### LINKED ISSUE #66: Fix `ret_gains` behavior
`ret_gains` is a new kwarg for the `add_` methods of the `Simulator` class, introduced in #63. It currently returns the ratio of the new data array (post `add_` method) to the old data array, but this isn't exactly the desired behavior. In particular, we would like the per-antenna gains returned as a dictionary (this is how they are handled under the hood in `sigchain`).

### LINKED ISSUE #68: Improve `gen_sigchain_reflections`
Right now, it is difficult to generate cable reflections for an arbitrary array&mdash;doing so requires preparing a few extra arrays, which makes this function basically unusable from a configuration file. As a first fix, a new YAML tag can be made to handle this.

### LINKED ISSUE #69: Support Multiple Polarizations in Simulator/Tsky Objects
Update the YAML constructors, `Simulator` `add_` methods, and `Tsky` object to support multiple polarizations. Ensure that `Simulator.add_foregrounds` and the like iterate over polarizations appropriately.

### LINKED ISSUE #71: Improve Tutorial Notebooks
The tutorial notebooks could be improved a bit, especially the tutorial for `defaults`. These should be updated prior to merging in the version 1 release.

### LINKED ISSUE #93: Use `deprecation` package to issue deprecation warnings for functions

### LINKED ISSUE #96: chunk_sim_and_save method
Coming from the hera-validation effort, a function that chunks up a large simulation (in time) and saves it as a set of files is useful. Could add it to `hera_sim.io`, and add a thin wrapper to `SImulator`.

### LINKED ISSUE #97: Metadata adjustment/matching from one simulation file to another
This would do something like is being done in test 4.0.0 for validation: taking a simulation, rephasing to another simulation, and matching the metadata.

### LINKED ISSUE #98: Match antennas from one simulation to another file
Similar to what happened in test 4.0.0 for validation.

### LINKED ISSUE #101: Use pre-commit/flake/black
These are becoming standard across hera/rasg.

### LINKED ISSUE #108: Installation instructions are wrong, environment.yml has some issues.
The reference a non-existent conda yaml file: `conda env create -f travis-environment.yml`. Note that this same mistake is made in the contributing guidelines.

I presume you mean to refer to the file `environment.yml`, which seems to have some packages that you don't need (nose, future -- you only seem to support python 3) and it's missing `astropy`, which is required by pyuvdata, but since you're installing that via pip you may end up with a pip installation of astropy rather than a conda installation, which is definitely not optimal (an aside: yhy are you installing pyuvdata via pip rather than conda?). Also, the `environment.yml` file lists `healpy` and `astropy-healpix`, but only the latter appears in `setup.py` -- is `healpy` actually required? Same question for `numba` and `basemap`.

### LINKED ISSUE #109: No License
There's no license on this repo.

Note that we usually use a permissive open source license for HERA repos (e.g. BSD or MIT). But that may not be allowed if you are actually using healpy (which has a copy-left license).

### LINKED ISSUE #125: Refactor `utils`, add filter cached property to `Simulator`
Refactor the `utils.rough_<delay/fringe>_filter` functions so that a pre-computed filter may be used (rather than having a new filter be calculated each time) and update the `Simulator` class to keep a cache of delay/fringe filters relevant for the given setup. It may be best to leave this as an optional functionality, in case the simulation is running up against memory limits without a cache of filters (for example, in a HERA-350-like simulation). Backwards compatibility should be maintained when refactoring the filter functions.

### LINKED ISSUE #138: Improve readthedocs API reference
As we move to incorporating `typing` into writing functions, it would be useful to look into whether it's possible to make the documentation preserve line breaks from the code when it is automatically generated from the code. An example of a less-than-optimal documented signature and the actual signature is provided below.

signature_example_code
signature_example

r-pascua:
This PR introduces the new implementation of `hera_sim`. Most of the old functions that existed prior to this version 1 release are still present, but the way they are implemented has changed. At the time of opening, this PR still needs much work: documentation needs to be filled out for everything and discussions need to be had with regular users regarding the proposed new implementations, among other things.

Closes #71 
Closes #69 
Closes #68 
Closes #66 
Closes #65 
Closes #64 
Closes #55 
Closes #25 
Closes #24
Closes #109 
Closes #138
Closes #125 
Closes #108 
Closes #101 
Closes #98 
Closes #97 
Closes #96 
Closes #93 
Closes #58


bhazelton:
@r-pascua FYI, you need to have `closes` before each issue you want this PR to close. Currently it will only close the first one you list. You can tell what it's going to do by looking at the "linked issues" area on the right sidebar.

r-pascua:
Gotcha, thanks for the info! I'll update it.

r-pascua:
@steven-murray It looks like I need an approving review to merge this. I think I'm done changing things for a while, so we should get this merged when you have a second to click the shiny "approve" button. There are still some things I have small issues with, but I'm happy to address them later.

On another note, this is a huge number of commits. I feel like a rebase would be nice, but I'm honestly terrified of attempting one&mdash;I feel like I'd end up losing a day only to admit defeat at the end of it all. Do you have thoughts on this?

Also, we need to make sure to change `master` to `main` after we get this merged, per #129.

steven-murray (approved):
Well. Here we are. This is a significant piece of work! Well done @r-pascua !

PR #78, opened 2020-01-10
Fix 77

### LINKED ISSUE #77: VisibilitySimulator shouldn't need sky_intensity or point_sources if its in the obsparams
If there's a valid reference to a catalog in `obsparams`, we shouldn't need to pass in `point_sources`.

steven-murray:
Fixes #77 

In addition, fixes a bug with `HealVis` in which the order of setting the beams matters (since `HealVis` has to overwrite `pyuvsim`'s analytic beam).


hera_sim/tests/test_vis.py
```diff
@@ -45,6 +46,72 @@ def uvdataJD():
...
+        sky_freqs=np.unique(uvdata.freq_array),
+        point_source_flux=point_source_flux,
+        point_source_pos=point_source_pos,
+        nside=2**4
+    )
+
+    assert len(hv.beams) == 1
+    assert isinstance(hv.beams[0], healvis.beam_model.AnalyticBeam)
+
+    # Now try creating with an obsparam file
```
mkolopanis: These are different enough I would argue they should be different tests.

hera_sim/visibilities/simulators.py
```diff
@@ -93,6 +94,14 @@ def __init__(self, obsparams=None, uvdata=None, sky_freqs=None,
              self.beams,
              self.beam_ids) = initialize_uvdata_from_params(obsparams)
 
+            if point_source_pos is None:
+                try:
+                    catalog = initialize_catalog_from_params(obsparams)[0]
+                    point_source_pos = np.array([catalog['ra_j2000'], catalog['dec_j2000']]).T * np.pi/180.
+                    point_source_flux = np.atleast_2d(catalog['flux_density_I'])
+                except KeyError:
+                    pass
```
mkolopanis: I'm not 100% sure why we're okay with just `pass`ing if the coordinate system is what we expected or missing a key :thinking: but I'm sure there is a good reason to do so.

steven-murray: It'll raise a `KeyError` if there is no `catalog` entry in the `obsparams` file. This is fine, because we allow the catalog to be passed directly to the class as well.

hera_sim/tests/test_vis.py
steven-murray: yeah 'k/

hera_sim/visibilities/simulators.py
r-pascua: I think it might be helpful to put a brief comment on line 98, something like `# setup point sources from catalog if provided`, with another comment explaining that the `KeyError` is raised if a catalog is not specified in the `obsparams`. I'm convinced that it's ok to `pass` here, but it took a bit of digging to convince myself of that.

tmp.csv
```diff
@@ -0,0 +1,6 @@
+Name     Number   BeamID   E          N          U
```
r-pascua: Why did this file make its way into the repo?

hera_sim/tests/test_vis.py
```diff
@@ -102,7 +173,8 @@ def test_shapes(uvdata, simulator):
         sky_intensity=I_sky,
     )
 
-    assert v.simulate().shape == (NTIMES*len(uvdata.get_antpairs()), 1, NFREQ, 1) # len(uvdata.get_antpairs()) = number of baselines
+    assert v.simulate().shape == (NTIMES * len(uvdata.get_antpairs()), 1, NFREQ,
```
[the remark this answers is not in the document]
steven-murray: Yup, good catch (although not really part of this PR)

tmp.csv
steven-murray: Woops, my mistake. I've  `git rm`'d it now.

PR #79, opened 2020-01-17
Fix output shape of resample_Tsky with callable input

steven-murray:
This fixes the shape of the output of `resample_Tsky` when given callable input (i.e. a spline). To be honest, I'm not sure why the test was passing previously...

### DIFF
--- a/hera_sim/noise.py
+++ b/hera_sim/noise.py
@@ -115,11 +115,15 @@ def resample_Tsky(fqs, lsts, Tsky_mdl=None, Tsky=180.0, mfreq=0.18, index=-2.5):
         tsky = Tsky_mdl(lsts, fqs)  # support an interpolation object
 
         if tsky.shape != (len(lsts), len(fqs)):
-            warnings.warn(
-                "Tsky_mdl should be a callable that takes (lsts, fqs) and returns an"
-                "array with shape (nlsts, nfqs). Note that interp2d objects do"
-                "*not* return this shape! Transposing array...")
-            tsky = tsky.T
+            msg = ("Tsky_mdl should be a callable that takes (lsts, fqs) and returns an"
+                    "array with shape (nlsts, nfqs). Note that interp2d objects do"
+                    "*not* return this shape! Transposing array...")
+
+            if tsky.shape == (len(fqs), len(lsts)):
+                warnings.warn(msg)
+                tsky = tsky.T
+            else:
+                raise ValueError(msg)
 
     else:
         tsky = Tsky * (fqs / mfreq) ** index  # default to a scalar
--- a/hera_sim/noise.py
+++ b/hera_sim/noise.py
@@ -112,7 +112,15 @@ def resample_Tsky(fqs, lsts, Tsky_mdl=None, Tsky=180.0, mfreq=0.18, index=-2.5):
             sky temperature vs. time and frequency
     """
     if Tsky_mdl is not None:
-        tsky = Tsky_mdl(lsts, fqs).T  # support an interpolation object
+        tsky = Tsky_mdl(lsts, fqs)  # support an interpolation object
+
+        if tsky.shape != (len(lsts), len(fqs)):
+            warnings.warn(
+                "Tsky_mdl should be a callable that takes (lsts, fqs) and returns an"
+                "array with shape (nlsts, nfqs). Note that interp2d objects do"
+                "*not* return this shape! Transposing array...")
+            tsky = tsky.T
+
     else:
         tsky = Tsky * (fqs / mfreq) ** index  # default to a scalar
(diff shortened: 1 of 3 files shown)


r-pascua:
I'm a little confused. Why would you expect `Tsky_mdl(lsts, fqs).T` to have shape `(NTIMES, NFREQS)`? I would expect an interpolation object `Tsky_mdl` to produce an output of shape `(NTIMES, NFREQS)` If called as `Tsky_mdl(lsts, fqs)`. Does this new test pass on your local setup? I can't tell on the PR because all of the builds failed due to the test modules not being recognized as tests.

lwhitler:
We were having an issue with using the 2D interpolation in `scipy` for `Tsky_mdl`, specifically. Creating the interpolation function requires calling `scipy.interpolate.interp2d(freqs, lsts, data)`, despite the data having dimensions of `(NTIMES, NFREQS)`—that seems to be a feature of `scipy`. The other way around (`scipy.interpolate.interp2d(lsts, freqs, data)`) throws this: `ValueError: When on a regular grid with x.size = m and y.size = n, if z.ndim == 2, then z must have shape (n, m)`. Later calling the interpolation function returns an array with dimensions `(NFREQS, NTIMES)`.

```
import numpy as np
from scipy.interpolate import interp2d

lsts = np.linspace(0, 0.05, 60)
freqs = np.linspace(1e8, 2e8, 1024)
data = np.random.rand(len(lsts), len(freqs))

interp_f = interp2d(freqs, lsts, data)
data_new = interp_f(lsts, freqs)

print('Shape of LST array: {}'.format(lsts.shape))
print('Shape of frequency array: {}'.format(freqs.shape))
print('Shape of original array: {}'.format(data.shape))
print('Shape of interpolated array: {}'.format(data_new.shape))
```

We also tried transposing the data when we created the interpolation function(`scipy.interpolate.interp2d(lsts, freqs, data.T)`) and that had the same problem.

r-pascua:
Aha! Your problem is due to the interpolation function you're using. If you look under the hood, you'll see that `interp2d` is kind of... weird. The gist of it is that by design it returns an array whose shape is the transpose of the shape of the passed array. I'm not sure why they do it this way, but that's the way it's done. The thing is that for interpolation objects like `RectBivariateSpline`, the shape is actually preserved--so for the `HERA_Tsky_mdl`, which is constructed using a `RectBivariateSpline`, this returns something with the correct shape. To convince yourself of this, rewrite the code block you shared using `RectBivariateSpline` in place of `interp2d`. As an extra note, for the case provided above, the domain is a rectangular grid, so `RectBivariateSpline` is actually the preferred interpolation method (the documentation for `interp2d` supports this claim). 

Moving forward, I would suggest doing the following:
Remove the transpose that was added, but add a check after resampling the sky model, something like
```
if tsky.shape != (lsts.size, fqs.size):
    if tsky.shape == (fqs.size, lsts.size):
        warnings.warn(
            "The interpolation object specified seems to return the transpose of the "
            "resampled sky model. Returning the transpose of the interpolator output."
        )
        tsky = tsky.T
    else:
        raise TypeError(
            "Resampled sky model has the wrong shape. Check your sky model."
        )
```
Now, this isn't super robust (it will throw weird errors if a non-callable object is passed as `Tsky_mdl`), but that's fine.

@steven-murray to answer why this test wasn't failing before: `HERA_Tsky_mdl` is a `RectBivariateSpline` instance created to be called as `HERA_Tsky_mdl['xx'](lsts, freqs)` (or `'yy'` for the polarization), at least in any version of `hera_sim` after I added the `interpolators.py` module.

steven-murray:
Thanks @r-pascua, that's a good suggestion. I've made a commit to that effect. So this should now work for `interp2d`-callables and "normal" callables.

hera_sim/noise.py
```diff
@@ -113,6 +113,14 @@ def resample_Tsky(fqs, lsts, Tsky_mdl=None, Tsky=180.0, mfreq=0.18, index=-2.5):
     """
     if Tsky_mdl is not None:
         tsky = Tsky_mdl(lsts, fqs)  # support an interpolation object
+
+        if tsky.shape != (len(lsts), len(fqs)):
```
lwhitler: I think this is assuming that if tsky doesn't have dimensions of `(NTIMES, NFREQS)`, it has dimensions of `(NFREQS, NTIMES)` so transposing it will give sensible output--should we also worry about when tsky just has nonsensical dimensions (i.e., neither `(NTIMES, NFREQS)` nor `(NFREQS, NTIMES)`)?

steven-murray: Good catch. Check the newest commit -- raises an error if it's a weird shape, but a warning if it can fix it.

PR #90, opened 2020-04-02
Add ability to use autos for creating noise

steven-murray:
This PR adds the ability to add noise directly based on the autocorrelations of a single antenna.


steven-murray:
This has already been addressed in v1.0.0

PR #91, opened 2020-04-12
Added GPU acceleration to vis_cpu

### LINKED ISSUE #92: test_vis.test_healvis_beam_obsparams does not pass
ValueError: no field of name flux_density_I. To me there is instead a field of name "I"? Just wanted to confirm it was not just a problem on my end before fixing it since I don't know what change would have caused this.

AaronParsons:
This pull request is pending the merge of vis_gpu support into hera_gpu master. It slightly modifies the vis_cpu function to match the new treatment of data types in vis_gpu, then adds a 'use_gpu" flag to visCPU to enable GPU acceleration.

Results are impressive, as tested on my Thinkpad P1 gen2 laptop with NVIDIA Quadro T2000.

In visibility_simulator.ipynb, jumping to the HERA simulation section and setting nbase = 9 (for a healpix nside of 512 for a 37-element array), healvis estimates:
Finished: 1, Elapsed 5.61min, Remain 4.398hour, MaxRSS 1.258044GB

by setting use_gpu=True in VisCPU for the same simulation, it *completes* in 28 seconds.

All other aspects of VisCPU(..., use_gpu=True) behave identically in the notebook.

Fixes #92.

### DIFF
--- a/hera_sim/visibilities/vis_cpu.py
+++ b/hera_sim/visibilities/vis_cpu.py
@@ -46,8 +46,10 @@ class VisCPU(VisibilitySimulator):
             try:
                 from hera_gpu.vis import vis_gpu
                 self._vis_cpu = vis_gpu
-            except(ImportError):
-                raise ImportError('GPU acceleration requires hera_gpu (https://github.com/HERA-Team/hera_gpu).')
+            except ImportError:
+                raise ImportError(
+                    'GPU acceleration requires hera_gpu (`pip install hera_sim[gpu]`).'
+                )
         else:
             self._vis_cpu = vis_cpu
         self.bm_pix = bm_pix
--- a/hera_sim/visibilities/vis_cpu.py
+++ b/hera_sim/visibilities/vis_cpu.py
@@ -43,8 +43,11 @@ class VisCPU(VisibilitySimulator):
             self._complex_dtype = np.complex128
 
         if use_gpu:
-            from hera_gpu.vis import vis_gpu
-            self._vis_cpu = vis_gpu
+            try:
+                from hera_gpu.vis import vis_gpu
+                self._vis_cpu = vis_gpu
+            except(ImportError):
+                raise ImportError('GPU acceleration requires hera_gpu (https://github.com/HERA-Team/hera_gpu).')
         else:
             self._vis_cpu = vis_cpu
         self.bm_pix = bm_pix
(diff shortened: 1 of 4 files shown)


hera_sim/visibilities/vis_cpu.py
```diff
@@ -34,8 +33,20 @@ def __init__(self, bm_pix=100, real_dtype=np.float32,
...
+        self._precision = precision
+        if precision == 1:
+            self._real_dtype = np.float32
+            self._complex_dtype = np.complex64
+        else:
+            self._real_dtype = np.float64
+            self._complex_dtype = np.complex128
+
+        if use_gpu:
+            from hera_gpu.vis import vis_gpu
```
steven-murray, line 46: Given that I would expect `hera_gpu` to be an optional depency, we should wrap this import with a `try: except ImportError:` and raise with some instruction on how to install GPU support (I expect it will be something like `pip install .[gpu]`.

hera_sim/visibilities/vis_cpu.py
```diff
@@ -34,8 +33,20 @@ def __init__(self, bm_pix=100, real_dtype=np.float32,
...
-        self._real_dtype = real_dtype
-        self._complex_dtype = complex_dtype
+        assert precision in (1,2)
+        self._precision = precision
+        if precision == 1:
+            self._real_dtype = np.float32
+            self._complex_dtype = np.complex64
+        else:
+            self._real_dtype = np.float64
+            self._complex_dtype = np.complex128
```
steven-murray, line 43: Are these definitions used at all? Since you've changed the signature of `vis_cpu` to also take the `precision` parameter (which I think is much cleaner!), you don't need these dtypes at all.

hera_sim/visibilities/vis_cpu.py
AaronParsons: Added a generic statement. Feel free to tweak if desired.

hera_sim/visibilities/vis_cpu.py
AaronParsons: Thw class builds arrays that get passed to vis_cpu/vis_gpu, and it's more efficient for them to be built in the correct datatype, so I think these should stay.

AaronParsons:
@steven-murray vis_gpu was merged into hera_gpu. I would like the support for it to get activated here, but I am not seeing where vis_cpu gets tested to address your comments. Given my extremely low bandwidth these days, do you think you could give me a hand? It should literally just be a copy-paste with use_gpu=True.

steven-murray:
@AaronParsons. Yup I can probably finish this PR off on Thursday. Tests will be basically a one-liner.

PR #102, opened 2020-05-22
Infrastructure

steven-murray:
This adds:

- black formatting
- flake8 checks
- Github Actions CI
- setuptools_scm versioning
- declarative setup.cfg
- codecov


docs/conf.py
```diff
@@ -16,32 +16,39 @@
...
-project = 'hera_sim'
-copyright = '2018, HERA-Team'
-author = 'HERA-Team'
+project = "hera_sim"
+copyright = "2020, HERA-Team"
+author = "HERA-Team"
 
 # The short X.Y version
-version = ''
+version = ""
```
r-pascua, line 49: Do we want to update this to be 1.0?

docs/installation.rst
```diff
@@ -5,13 +5,13 @@ Installation
...
 Requires:
- * `numpy`
- * `scipy`
- * `aipy`
- * `hera_cal` (which requires `h5py`)
- * `pyuvdata`
+ * ``numpy``
+ * ``scipy``
+ * ``aipy``
+ * ``hera_cal`` (which requires ``h5py``)
```
r-pascua, line 11: It looks like the only place `hera_cal` is used is in `vis.py`, and it's just importing `utils.split_bl` from `hera_cal` inside the `sim_red_data` function. I don't think this PR is the place to deal with this, but I think we may want to think a bit about whether it's sensible to keep `hera_cal` as a requirement given its extremely limited use.

docs/installation.rst
```diff
@@ -5,13 +5,13 @@ Installation
 Requirements
 ------------
 Requires:
```
r-pascua, line 7: I'm not seeing `astropy` on this list, but it should definitely be there.

docs/notes_for_developers.rst
```diff
@@ -1,12 +1,10 @@
...
+Developing ``hera_sim``
+=======================
 
-`hera_sim` broadly follows the best-practices laid out in XXX.
+``hera_sim`` broadly follows the best-practices laid out in XXX.
 
 .. todo:: where is that best-practices doc?
 
 
 All docstrings should be written in `Google docstring format <https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html>`_.
```
r-pascua, line 10: Not the place to address it, but just want to note that I thought we switched to numpy docstring format (and so this should be changed in the future).

docs/conf.py
steven-murray: Actually, the version should be gotten automatically... I'll update this

steven-murray: This should be done now.

docs/installation.rst
steven-murray: I've made a new extra requires for `cal` which installs `hera_cal`, and otherwise removed it from installation instructions (it wasn't being installed automatically anyway).

docs/installation.rst
steven-murray: This list/instructions should definitely be updated -- it doesn't actually match the requirements in `setup.cfg`. I think I'll leave that to another PR.

docs/notes_for_developers.rst
steven-murray: Yup, definitely. In the PR that we do that, we should also change this.

PR #105, opened 2020-06-16
CLI revamp

r-pascua:
This PR implements an overhaul to the command-line interface. We drop the use of `click` and instead use the `argparse` module for handling the CLI. The CLI itself is moved into a new `scripts` directory and is titled `simulate.py`. Some extra infrastructure may still need to be added, and the CLI has yet to be rigorously tested, though it works using the template configuration file (which tests most every feature of the CLI).


r-pascua:
It looks like I've fixed all of the test failures related to code I wrote (modulo linting/pre-commit stuff). The only failing test now is the `healvis`-related issue. I've glanced at the output from the linting/pre-commit step, and I don't understand what the failure modes mean, so I could use some pointers on that.

Do we want to try to implement some sort of unit test for the CLI? I *really* don't like the issues around testing CLIs, though I do think it's important to do something to make sure that the CLI doesn't break from new modifications/additions...

steven-murray:
Wow, this is a lot of work!

To fix the linting step, you probably just need to do

```bash
pre-commit autoupdate
pre-commit run -a
```
If the second step fails, just run it again. If it still fails, let me know what it's saying.

I do think we need to test the CLI, at least to a basic level, otherwise it will pretty quickly grow stale.
I would just do stuff like give it some inputs, and test that the files it's supposed to create are there.

hera_sim/cli_utils.py
```diff
@@ -0,0 +1,136 @@
...
+    filename, 
+    sim=None, 
+    freqs=None, 
+    times=None, 
+    clobber=False
+):
+    """Write gains to disk as a calfits file."""
+    from hera_cal.io import write_cal
+    if sim is not None:
+        if isinstance(sim, Simulator):
```
steven-murray, line 64: I'm now slightly wary of passing around the `Simulator` class and modifying its `data` object, given our troubles in validation 4.0. Not sure it would be an issue here particularly, but we should keep an eye out for it.

hera_sim/cli_utils.py
```diff
@@ -0,0 +1,136 @@
...
+def write_vis(ref_sim, vis, filename, save_format="uvh5", bda_params=None, **kwargs):
+    """Write a simulated visibility to disk."""
+    tmp_sim = copy.deepcopy(ref_sim)
+    if not isinstance(tmp_sim, Simulator):
+        tmp_sim = Simulator(data=tmp_sim)
+
+    tmp_sim.data.data_array = vis
+    if bda_params:
+        # This should be installed if this clause is executed.
+        from bda import bda_tools
```
steven-murray, line 90: Make sure this is an `extra_requires`

hera_sim/cli_utils.py
```diff
@@ -0,0 +1,136 @@
...
+    write_cal(filename, gains, freqs, times, overwrite=clobber, return_uvc=False)
+
+def write_vis(ref_sim, vis, filename, save_format="uvh5", bda_params=None, **kwargs):
+    """Write a simulated visibility to disk."""
+    tmp_sim = copy.deepcopy(ref_sim)
+    if not isinstance(tmp_sim, Simulator):
+        tmp_sim = Simulator(data=tmp_sim)
+
+    tmp_sim.data.data_array = vis
+    if bda_params:
```
steven-murray, line 88: I wonder if the bda shouldn't be applied outside this function -- having bda applied inside a function called `write_vis` kind of goes against the "single responsibility principle".

hera_sim/simulate.py
```diff
@@ -57,26 +57,31 @@ def __init__(self, data=None, defaults_config=None, **kwargs):
         # actually initialize the UVData object stored in self.data
         self._initialize_data(data, **kwargs)
 
-    @cached_property
+    @property
     def antpos(self):
         # TODO: docstring
```
steven-murray, line 62: TODO

hera_sim/simulate.py
```diff
@@ -57,26 +57,31 @@ def __init__(self, data=None, defaults_config=None, **kwargs):
...
         """
         """
         antpos, ants = self.data.get_ENU_antpos(pick_data_ants=True)
         return dict(zip(ants, antpos))
 
-    @cached_property
+    @property
     def lsts(self):
```
steven-murray, line 69: TODO

hera_sim/simulate.py
```diff
@@ -57,26 +57,31 @@ def __init__(self, data=None, defaults_config=None, **kwargs):
...
         # TODO: docstring
         return np.unique(self.data.lst_array)
 
-    @cached_property
+    @property
     def freqs(self):
         # TODO: docstring
```
steven-murray, line 75: TODO is done!

hera_sim/simulate.py
```diff
@@ -57,26 +57,31 @@ def __init__(self, data=None, defaults_config=None, **kwargs):
...
         """Frequencies in GHz
         """
         return np.unique(self.data.freq_array) / 1e9
 
+    @property
+    def times(self):
+        """Return unique simulation times."""
+        return np.unique(self.data.time_array)
+
     # XXX begin methods intended for user interaction XXX
```
steven-murray, line 85: remove this comment

hera_sim/simulate.py
```diff
@@ -276,15 +282,34 @@ def write(self, filename, save_format="uvh5", save_seeds=True, **kwargs):
         # TODO: docstring
```
steven-murray, line 282: TODO

hera_sim/simulate.py
```diff
@@ -276,15 +282,34 @@ def write(self, filename, save_format="uvh5", save_seeds=True, **kwargs):
...
         try:
             getattr(self.data, "write_%s" % save_format)(filename, **kwargs)
         except AttributeError:
             msg = "The save_format must correspond to a write method in UVData."
             raise ValueError(msg)
-        if save_seeds:
-            seed_file = os.path.splitext(filename)[0] + "_seeds"
-            np.save(seed_file, self._seeds)
 
+    # XXX with the new version of the CLI, this should not need to be wrapped
```
steven-murray, line 311: Is this true? Can it be tested?

hera_sim/simulate.py
```diff
@@ -904,7 +930,7 @@ def _update_history(self, model, **kwargs):
         msg = "hera_sim v{version}: Added {component} using kwargs:\n"
         if defaults._override_defaults:
             kwargs["defaults"] = defaults._config_name
-        for param, value in kwargs.items():
+        for param, value in defaults._unpack_dict(kwargs).items():
             msg += "{param} = {value}\n".format(param=param, value=value)
```
steven-murray, line 934: can use f-string formatting

scripts/simulate.py
```diff
@@ -0,0 +1,192 @@
+"""
+CLI for hera_sim
```
steven-murray, line 2: Be a little bit more expansive in the description -- we might yet have other CLI scripts :-)

scripts/simulate.py
```diff
@@ -0,0 +1,192 @@
...
+import hera_sim
+from hera_sim import cli_utils
+from pyuvsim.simsetup import _parse_layout_csv
+from astropy.coordinates import Angle
+from astropy import units
+try:
+    import bda
+    from bda import bda_tools
+except ImportError:
+    warnings.warn("bda failed to import. Baseline-dependent averaging not available.")
```
steven-murray, line 20: This might get annoying -- may just want to use an appropriate error if/when bda is actually _used_.

scripts/simulate.py
```diff
@@ -0,0 +1,192 @@
...
+)
+parser.add_argument(
+    "--clobber",
+    default=None,  # To allow the user to specify clobber in the config file.
+    action="store_true",
+    help="Overwrite existing files in case of name conflicts."
+)
+args = parser.parse_args()
+
+if args.verbose:
```
steven-murray, line 56: Shouldn't everything from here down be in a `if __name__ == "__main__"`?

scripts/simulate.py
```diff
@@ -0,0 +1,192 @@
...
+    print("Preparing simulation...")
+
+instrument_parameters = {}
+defaults = config.get("defaults", None)
+if defaults is not None:
+    # TODO: update the defaults module to handle validating the choice of defaults
+    if not isinstance(defaults, str):
+        msg = "Defaults must be specified using a string. The string may be either "
```
steven-murray, line 90: You can literally do `ValueError("Defaults must be...")` and put new lines between each line -- as long as it's in brackets, strings by default concatenate between lines.

scripts/simulate.py
```diff
@@ -0,0 +1,192 @@
...
+            continue
+
+        simulation_parameters[this_component] = settings
+
+if args.verbose:
+    print("Running simulation...")
+
+# XXX it may be simpler to just use sim.run_sim for this, and if the user 
+# wants to save the components, then we can loop over the items in 
+# simulation_parameters and just do sim.get
```
steven-murray, line 151: Not sure I understand this comment, but thought I should note it here.

config_examples/template_config.yaml
```diff
@@ -17,9 +17,10 @@ bda:
         units: 's'
 filing:
     outdir: '.'
-    outfile_name: 'quick_and_dirty_sim'
+    outfile_name: 'quick_and_dirty_sim.uvh5'
     output_format: 'uvh5'
     clobber: True
+    # TODO: remove this after updating how writing seeds is handled
```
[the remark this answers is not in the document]
r-pascua: Sort of. The current state of it is that I have a `save_seeds` parameter in `Simulator.write`, and that's where putting the seeds in a disk-writeable format (modulo `miriad`-type files) is handled. I'm thinking the best way to improve this is to just run a step called something like `_update_seed_tracking` after each new component is added to a simulation, then it's not even a question of whether or not to save the seeds. Granted, this needs some fine tuning--last night was a sort of first-pass at wrapping up the last outstanding issues before making the PR.

config_examples/template_config.yaml
```diff
@@ -54,32 +55,37 @@ telescope:
...
 systematics:
     rfi:
         # see hera_sim.rfi documentation for details on parameter names
         rfi_stations:
+            seed: once
             stations: !!null
         rfi_impulse:
-            chance: 0.001
-            strength: 20.0
+            impulse_chance: 0.001
```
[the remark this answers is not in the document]
r-pascua: Ah, so this is because of modifications I made to the `rfi` module to keep some of the `defaults` logic simple. Basically, `defaults` unravels whatever configuration it's given so that any level of dictionary nesting is removed--so having multiple instances of `chance` or `std` (rather than differentiating things as, e.g., `scatter_chance` and `impulse_chance`) would break `defaults` handling of `rfi` component parameters.

hera_sim/cli_utils.py
r-pascua: Agreed. It's not an issue here, since it's just used to pull required metadata if not provided by the other keywords, but the issues in the 4.0 test definitely warrant the wariness.

hera_sim/cli_utils.py
r-pascua: I think it probably should be applied outside this function... I'll update it.

hera_sim/simulate.py
r-pascua: Planning on leaving this for a PR dedicated to updating documentation, unless you're strongly against this.

hera_sim/simulate.py
r-pascua: Same comment as above

hera_sim/simulate.py
r-pascua: Good catch!

hera_sim/simulate.py
r-pascua: OK. There's a similar comment further down; I'll remove that too.

hera_sim/simulate.py
r-pascua: Same comment as previous comment about addressing unattended documentation TODOs.

hera_sim/simulate.py
r-pascua: It's definitely true now. The previous version of the CLI basically parsed the passed configuration file to generate a configuration to be passed to `Simulator.run_sim`, so the generator-like behavior was needed in order to actually extract the individual data products. The new version of the CLI just loops over the parsed configuration and applies effects using the `Simulator.add` method.

hera_sim/simulate.py
r-pascua: Yep. This is *old* code, and I didn't know about f-strings at the time. There are a few other instances of this around the code... I was thinking of getting all of these stylistic issues resolved in one fell swoop in a `code_beautification` PR.

scripts/simulate.py
r-pascua: Agreed, good catch.

scripts/simulate.py
r-pascua: I think I can maybe just remove the `warnings.warn(...)` statement, since there is a step in the CLI that checks to make sure `bda` is installed if `bda_params` is specified in the configuration and errors if it isn't.

scripts/simulate.py
r-pascua: I think that is best practice, yes. I'll make the update.

scripts/simulate.py
r-pascua: Yep... I think this was more of a "this looks nicer to me" thing? I'm not sure though. I seem to recall having an issue with something involving breaking up long strings across lines, but that may have been with `warnings.warn`, rather than any built-in exceptions.

scripts/simulate.py
r-pascua: I think I'm just going to scrap the comment. The `get` method re-runs the simulation step, so it would make the CLI significantly slower when running the CLI with the `--save_all` flag.

scripts/simulate.py
r-pascua: This value-checking block has been removed, since it is handled in `cli_utils.validate_config`.

hera_sim/cli_utils.py
r-pascua: I ended up just removing this function altogether and put the logic for handling it into the CLI. I think it's better done that way, since there will be some complications with writing stuff to disk when BDA is being applied, both for gain-like and visibility-like terms. I think the implementation that's there now should appropriately handle visibility-like terms--I'm punting the handling of gain-like terms to when I add support for time-varying gains.

hera_sim/cli_utils.py
r-pascua: Looks like this has already been done, assuming I'm looking in the right place (`setup.cfg` in the top-level directory).

config_examples/template_config.yaml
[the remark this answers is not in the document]
r-pascua: No longer a TODO.

hera_sim/cli_utils.py
```diff
@@ -0,0 +1,136 @@
...
+
+def write_calfits(
+    gains, 
+    filename, 
+    sim=None, 
+    freqs=None, 
+    times=None, 
+    clobber=False
+):
+    """Write gains to disk as a calfits file."""
```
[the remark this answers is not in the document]
r-pascua: Updated documentation for this.

hera_sim/simulate.py
r-pascua: So I updated this comment a bit with something I thought was worth considering. The `_generator_to_list` decorator allows the user to receive a list with all (or a subset of) the different simulation components... this might be a handy feature. I kind of want to scrap it and require that the user uses `Simulator.get` to retrieve simulation components, especially because the implementation used here is prone to the return value being confusing.

Do you have any thoughts on this?

hera_sim/simulate.py
r-pascua: Updated this to use f-strings here, but there are plenty of other places where this needs to be done... hence the idea of making a `code_beautification` PR.

scripts/simulate.py
r-pascua: Updated the description. Lmk if you think I should make any changes.

config_examples/template_config.yaml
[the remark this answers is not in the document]
steven-murray: On the face of it this seems a bit brittle -- what if there are other modules in the future that also require this kind of handling? But I'm sure you have good reason for it, and there might not be a "better" solution.

hera_sim/cli_utils.py
steven-murray: Yup, cool.

hera_sim/cli_utils.py
steven-murray: Cool, sounds good!

hera_sim/simulate.py
steven-murray: Yup, that's fine.

hera_sim/simulate.py
steven-murray: "There should be one-- and preferably only one --obvious way to do it."

I think making the user be explicit and use `Simulator.get` is entirely reasonable, unless that can get non-obvious (perhaps if they potentially don't know what they've simulated?).

scripts/simulate.py
steven-murray: It's now `"""Re-imagining of the simulation module."""` right? I might go for something like "Simulate post-RIME effects via CLI"?

config_examples/template_config.yaml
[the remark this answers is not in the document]
r-pascua: It's been a while since I wrote the `defaults` module, but I believe that the motivation here was to keep the code relatively simple and to make the API for setting new defaults as simple as possible. To be honest, I'm not really sure what the *right* thing to do is... implementing package-wide, dynamic defaults handling is hard, so it's entirely possible that I haven't converged on the "correct" solution.

hera_sim/simulate.py
r-pascua: After thinking about this in the context of how I've used `hera_sim` for validation tests, I think I would rather scrap the option to have `Simulator.run_sim` return anything and require that end users who want particular simulation components either explicitly simulate effects one-by-one (and use the `ret_vis` keyword) or run the full simulation and use `Simulator.get`.

scripts/simulate.py
r-pascua: Ah, no, that's currently the docstring for the `simulate` module where the `Simulator` is defined. I suppose it's bad practice to name a script the same thing as one of the modules. The new docstring for the CLI is much more descriptive than just a one-liner. (I'll definitely need to update the `simulate` module's docstring at some point, though.)

config_examples/template_config.yaml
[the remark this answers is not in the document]
steven-murray: That sounds like the right goal haha. Let's go with this, and hopefully if issues arise we can tweak as we go.

steven-murray:
Looks like we just need to test  bit more of the diff.

r-pascua:
Cool, I'll get on updating the tests to increase the diff coverage. The question still remains of how to test the CLI though... Do you have any ideas or references regarding best practices for unit tests of CLIs?

steven-murray:
Honestly I think it's really functional tests you want to do at first -- just checking that the program runs as a whole (for small simulations!), and doesn't break with different argument combinations.

There's this: https://pypi.org/project/pytest-console-scripts/ that might make it a little easier. 

Otherwise I think pytest has some builtin functionality for capturing stdout etc which you can check. And check for names of produced files.

PR #112, opened 2020-07-02
changed recarray column name

aelanman:
A somewhat recent change to pyradiosky broke this call -- when a SkyModel is converted to a recarray, the flux data was being kept in a single column called `flux_density`. That has been changed -- separate columns for Stokes I, Q, U, and V are now included (with names simply `I`, `Q`, `U`, `V`).

We will be deprecating the recarray feature in the future, but for now this should fix test_vis.py.


aelanman:
Silly.... I forgot that Steven's PR already does this.... please disregard.

PR #113, opened 2020-07-03
Crosstalk spectrum

r-pascua:
This PR adds a new class in `sigchain` that allows the user to generate cross-coupling crosstalk spectra, as was done in the step-4 validation test.


r-pascua:
@nkern in this PR, I update the cross-coupling crosstalk simulation so that the crosstalk visibilities are symmetric in delay. I did this because that was what was done in the step-4 validation test; however, I notice that the old unit test for cross-coupling xtalk was written on the assumption that crosstalk shows up only at the delay specified, and not the conjugate delay. So, do you think this update is acceptable, or should I revert it so that it will only show up at the delay specified?

steven-murray (approved):
This all looks reasonable to me... sorry it must have slipped my attention for a good long while...

PR #114, opened 2020-07-04
Support time varying gains

r-pascua:
This PR ports over and extends the tools developed for making time varying gains as used in validation tests 2.1 and 4.0.


hera_sim/tests/test_sigchain.py
```diff
@@ -182,5 +185,339 @@ def test_cross_coupling_xtalk(self):
         )
 
 
+class TestTimeVariation(unittest.TestCase):
```
steven-murray, line 188: Don't use unittest stuff! Use `pytest` fixtures / parameterizations

r-pascua: Gotcha. I'll rewrite the new tests to use `pytest` only, and make the tests more atomic.

r-pascua: All of the tests for this module (including the old ones) have been updated to use `pytest` only. Some further consolidation can be done, and some tests may be able to be compactified by exploiting parameterization, but I'd like to leave that for the PR where I convert *all* of the tests to `pytest` only.

steven-murray: Excellent!

PR #115, opened 2020-07-07
Master merge

steven-murray:
Merges current master into v1.0.0


steven-murray:
I don't understand why the codecov status is not being reported. I wonder if it's the lingering continuous-integration/travis-ci/pr task, which will probably never run because travis is weird.

Perhaps we should just remove the whole travis thing, since github seems to be working well.

r-pascua:
I don't really understand why, but it looks like codecov thinks that substantially less of the `simulators` module has been covered? I guess this means the tests need to be updated before we can merge?

steven-murray:
I think it might be because we're now omitting tests from the coverage...

Yeah, we should try get the test coverage up again.

PR #126, opened 2021-02-03
Update tests

r-pascua:
This PR converts all tests to use `pytest` only. There are also some improvements to tests (increased coverage of some modules, less opaque tests, more rigorous tests) in various modules.

This is currently not complete, but there is a substantial amount of work that needs to be reviewed. The only remaining updates should be completion of rewriting the `Simulator` tests and perhaps better use of fixtures and parametrized tests.


hera_sim/tests/test_antpos.py
```diff
@@ -1,94 +1,97 @@
...
+            bl1 = ants[1] - ants[0]
+            bl2 = ants[hex_num] - ants[0]
+            expected_bl1 = sep * np.array([1, 0, 0])
+            expected_bl2 = sep * np.array([-0.5, -np.sqrt(3) / 2, 0])
+            assert np.allclose(bl1, expected_bl1)
+            assert np.allclose(bl2, expected_bl2)
+
+
+def test_hex_array_split_core_positions():
+    # This test is rather brittle and could be improved.
```
steven-murray, line 76: the only thing I can see to do is to set `sep` as a local variable

hera_sim/tests/test_antpos.py
```diff
@@ -1,94 +1,97 @@
...
+    bl1 = ants[3] - ants[7]
+    bl2 = ants[5] - ants[10]
+    expected_bl1 = 14.6 * np.array([0, 2 / np.sqrt(3), 0])
+    expected_bl2 = 14.6 * np.array([-1, 1 / np.sqrt(3), 0])
+    assert np.allclose(bl1, expected_bl1)
+    assert np.allclose(bl2, expected_bl2)
+
+
+def test_hex_array_HERA350_positions():
+    # This test is rather brittle and could be improved.
```
steven-murray, line 87: again with the `sep` variable

hera_sim/tests/test_cli_utils.py
```diff
@@ -37,6 +37,7 @@ def test_filing_params_parser():
         cli_utils.get_filing_params(bad_params)
 
 
+# TODO: break this up into multiple tests
```
steven-murray, line 40: yes, here would be a good place to user parameterization. 

Should be like:
```python
@pytest.mark.parameterize(["freq", "time", "array"], itertools.prod(freq_params, time_params, array_params))
def test_config_validation(freq, time,array, tmp_path):
    ...
```

hera_sim/tests/test_eor.py
```diff
@@ -1,50 +1,76 @@
...
+
+
+@pytest.fixture(scope="function")
+def scaled_eor(freqs, lsts, bl_vec, request):
+    np.random.seed(0)
+    if request.param == "auto":
+        bl_vec = np.array([0, 0, 0])
+    return eor.noiselike_eor(
+        lsts=lsts, freqs=freqs, bl_vec=bl_vec, eor_amp=1e-3, fringe_filter_type="tophat"
+    )
```
steven-murray, line 38: these can be combined with parameterization

hera_sim/tests/test_foregrounds.py
```diff
@@ -1,220 +1,138 @@
...
-        Tsky_mdl = noise.HERA_Tsky_mdl["xx"]
-        bl_vec = (0, 30.0)
-        fringe_filter_kwargs = {"fringe_filter_type": "tophat"}
-
-        # simulate the effect
+@pytest.mark.parametrize("model", ["pntsrc", "diffuse"])
+def test_foreground_shape(freqs, lsts, Tsky_mdl, omega_p, model):
+    bl_vec = [0, 0, 0]
+    if model == "pntsrc":
+        vis = foregrounds.pntsrc_foreground(lsts=lsts, freqs=freqs, bl_vec=bl_vec)
```
steven-murray, line 38: would it make sense to make pntsrc and diffuse vis fixtures? You use them in two different tests. Could scope them as "module" to make tests faster, since they're read-only.

hera_sim/tests/test_foregrounds.py
```diff
@@ -1,220 +1,138 @@
...
-        vis = foregrounds.diffuse_foreground(
-            lsts,
-            freqs,
-            bl_vec,
-            Tsky_mdl=Tsky_mdl,
-            omega_p=omega_p,
-            fringe_filter_kwargs=fringe_filter_kwargs,
+    elif model == "pntsrc":
+        # XXX point source foregrounds do not behave correctly in fringe-rate
+        # Is there a way to parametrize this that gives an xfail when point source
```
steven-murray, line 81: yes, see https://docs.pytest.org/en/stable/example/parametrize.html#parametrizing-conditional-raising

hera_sim/tests/test_foregrounds.py
```diff
@@ -1,220 +1,138 @@
...
-        # XXX this doesn't pass either!
-        np.testing.assert_allclose(vis, _vis.conj())
-
-
-if __name__ == "__main__":
-    unittest.main()
+def test_diffuse_foreground_exception_no_omega_p(freqs, lsts, Tsky_mdl):
+    bl_vec = [0, 0, 0]
+    with pytest.raises(ValueError) as err:
+        _ = foregrounds.diffuse_foreground(
```
steven-murray, line 135: don't need the _ here, just call the function without saving the return value.

hera_sim/tests/test_noise.py
```diff
@@ -1,117 +1,103 @@
...
-    unittest.main()
+@pytest.mark.parametrize("model", ["powerlaw", "HERA"])
+def test_resample_Tsky_freq_behavior(tsky_powerlaw, tsky_from_model, model):
+    tsky = tsky_powerlaw if model == "powerlaw" else tsky_from_model
+    assert all(np.all(tsky[:, 0] != tsky[:, i]) for i in range(1, tsky.shape[1]))
+
+
+@pytest.mark.parametrize("channel_width", [None, 1e6, 1e8])
+@pytest.mark.parametrize("integration_time", [None, 10.7, 33.7])
+@pytest.mark.parametrize("aspect", ["mean", "std"])
```
steven-murray, line 75: Since choosing between "aspects" here essentially branches the whole function, I think its cleaner just to make it two tests. You can make the `noise_Jy` a fixture so that it doesn't repeat.

hera_sim/tests/test_rfi.py
```diff
@@ -1,121 +1,161 @@
...
+    )
+    assert np.isclose(
+        np.mean(np.abs(rfi_vis).astype(bool)), expected_occupancy, atol=0.05, rtol=0,
+    )
+
+
+def test_rfi_dtv_warning_when_no_overlap(freqs, lsts):
+    dtv_band = (0.25, 0.37)
+    with pytest.warns(UserWarning) as warning:
+        _ = rfi.rfi_dtv(lsts, freqs, dtv_band=dtv_band)
```
steven-murray, line 129: no need for the _

steven-murray (commented):
Thanks @r-pascua this is awesome. I'd be essentially happy to merge as-is, but I noticed a few things we could probably make a little tighter.

hera_sim/interpolators.py
```diff
@@ -172,6 +172,8 @@ def _interpolator(self):
         # get the tsky data
         tsky_data = self.tsky[pol_index]
 
+        # XXX this isn't always what we want to do and will give unexpected behavior
+        # for sky models where the LST range only covers a subset of the sky.
         # do some wrapping in LST
         lsts = np.concatenate(
```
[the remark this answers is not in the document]
r-pascua: Done. Maybe a bit too verbose of a warning, but at least there is one for now.

hera_sim/rfi.py
```diff
@@ -285,14 +285,23 @@ def __call__(self, lsts, freqs, **kwargs):
...
         # define an iterator, just to keep things neat
+        df = np.mean(np.diff(freqs))
         dtv_iterator = zip(bands, dtv_chance, dtv_strength, dtv_std)
 
         # loop over the DTV bands, generating rfi where appropriate
         for band, chance, strength, std in dtv_iterator:
             # get the channels affected
-            ch1 = np.argwhere(band <= freqs)[0][0]
+            if any(np.isclose(band, freqs, 0.01 * df)):
+                ch1 = np.argwhere(np.isclose(band, freqs)).flatten()[0]
```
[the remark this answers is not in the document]
r-pascua: Done. Looks like I also made this error in the step finding `ch2`, and that has been fixed. I also think that the 0.01 * df is meant to be an absolute tolerance, so I've made that change as well.

hera_sim/simulate.py
```diff
@@ -30,6 +31,8 @@ def new_func(*args, **kwargs):
     return new_func
 
 
+# XXX: some of the code in here is pretty brittle and breaks (sometimes silently)
```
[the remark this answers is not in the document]
r-pascua: Fixed

hera_sim/simulate.py
```diff
@@ -229,14 +232,16 @@ def get(self, component, ant1=None, ant2=None, pol=None):
...
-            blt_inds = self.data.antpair2ind((ant1, ant2))
+            blt_inds = self.data.antpair2ind((ant1, ant2), ordered=False)
             if pol is None:
                 return data[blt_inds, 0, :, :]
             else:
                 pol_ind = self.data.get_pols().index(pol)
                 return data[blt_inds, 0, :, pol_ind]
         elif seed == "redundant":
-            if any([(ant2, ant1) == item for item in antpairpol_cache]):
+            # XXX: this will need to be modified when polarization handling is fixed
```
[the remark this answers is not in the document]
r-pascua: Fixed.

hera_sim/simulate.py
```diff
@@ -311,9 +316,7 @@ def run_sim(self, sim_file=None, **sim_params):
                 try:
                     sim_params = yaml.load(config.read(), Loader=yaml.FullLoader)
                 except Exception:
-                    print("The configuration file was not able to be loaded.")
-                    print("Please fix the file and try again.")
-                    sys.exit()
+                    sys.exit("The configuration file was not able to be loaded.")
```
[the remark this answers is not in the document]
r-pascua: We are now!

hera_sim/tests/test_foregrounds.py
r-pascua: Fixed all instances of something like this.

hera_sim/tests/test_antpos.py
r-pascua: I'm not sure if I quite understand what you mean. Would that entail something like perhaps parametrizing `sep` and checking that it works for different base separations? I think what I was getting at with the comment is that it seems like this test assumes a certain ordering of the antennas and basically checks that the ordering is obeyed and that the antennas are placed on a hexagonal grid (and I'd hope that it also tests that there are deviations from a perfect hexagonal grid by looking at baselines that use antennas in different sectors, but tbh I haven't dug quite that deep into validating these tests). I think this is fine as long as the algorithm for building the hex array doesn't change, but it kind of bothers me that it's so specific. Maybe it's not worth worrying over, though. What do you think?

hera_sim/tests/test_cli_utils.py
r-pascua: I've refactored this testing module.

hera_sim/tests/test_antpos.py
steven-murray: I think tests that are specific are fine -- if the code ever changes, the test will break. Then we'll be reminded that we broke API, and have to update documentation and version accordingly.

steven-murray:
I'm really not sure what's going on with the MacOS py38 build -- something about not installing numpy properly. @mkolopanis got any ideas?

mkolopanis:
Yeah this mac build thing is pretty scary. I have never personally seen it before. A quick run through your logs looks like it happens during the `healvis` install but it cloned two different repos without installing either before the error. Might want to poke the `healvis` dev team about updating their repo. lacks things like a pyproject.toml, and the setup.py does not specify every "required dependency" that is listed on its Readme. Now why would this cause this error? I don't know. unless pip does something crazy and makes a new temporary environment for every install it runs with a requirements.txt file.

mkolopanis:
unrelated, but in the RASG suite we have transitioned to using the `conda-incubator/setup-miniconda` action to well setup our miniconda environments in github actions. take a look [here](https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/8f60aa50fe2653f920f0e1c97cd2aca93e1d9862/.github/workflows/macosx_windows_ci.yaml#L24-L49) for an example. I personally find it makes the yamls a lot cleaner and then you don't have to worry so much about managing conda yourself.

hera_sim/tests/test_antpos.py
r-pascua: Ok, I can get on board with that. Do you think there are any changes I should make to this and the HERA-350 test?

hera_sim/tests/test_eor.py
r-pascua: err, I'm not sure what the right way to do this would be, since I use indirect parametrization of these fixtures in later tests. I'm also not doing any tests that only use `scaled_eor`--this is only used in checking that the visibilities scale appropriately (which is broken up into two tests since the way EoR is simulated for autos vs crosses gives different variances in the abs of the visibility).

steven-murray: maybe I misunderstand how indirect parameterization works. It's fine as-is, let's just leave it.

hera_sim/tests/test_foregrounds.py
r-pascua: eh, not really. maybe for the simple tests (shape testing and reality testing for autos), since those use the same set of visibilities? otherwise I think it would be a lot more of a headache to modify the tests to jive with making the foregrounds fixtures. the main issue I have here is that the tests on the crosses don't lend themselves well to reusing the visibilities: we only have two tests on the crosses as of now, and one of them requires simulating visibilities for two orientations (the given baseline and its conjugate) in the same test.

hera_sim/tests/test_foregrounds.py
r-pascua: argh, so this opened up a small can of worms. it looks like the point-source foregrounds get the fringe-rate behavior wrong for every baseline except the 100-m eastward baseline. so this isn't consistently "failing" as I had expected it to. I've made a little hack to deal with this, but I don't really like it.

hera_sim/tests/test_noise.py
r-pascua: I think that doing this will actually make the test code messier, because of these lines:
```
channel_width = channel_width or np.mean(np.diff(freqs)) * units.GHz.to("Hz")
integration_time = integration_time or units.day.to("s") / lsts.size
```
Based on my understanding of how to parametrize fixtures, there isn't an obviously neat way of determining the state of the `channel_width` and `integration_time` variables from the parametrized fixture. If you know of a way to do this, then I'd be happy to make the change with your help.

steven-murray (approved):
Thanks @r-pascua, this is excellent. I'll try figuring out the installation error, and we'll merge this one in.

PR #130, opened 2021-02-19
doc: a few small fixes to the readme for installation

### LINKED ISSUE #108: Installation instructions are wrong, environment.yml has some issues.

steven-murray:
Fixes #108

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,16 +28,19 @@ can handle natively:
 If you are creating a new development environment, consider using the included environment
 file:
 
-``$ conda env create -f travis-environment.yml``
+``$ conda env create -f ci/tests.yaml``
 
 This will create a fresh environment with all required dependencies, as well as those
-required for testing. If you also would like to compile the documentation, use
+required for testing. If you also would like to compile the documentation, use (after
+the above):
 
 ``$ pip install -r docs/requirements.txt``
 
 Finally, install `hera_sim` with
 
 ``$ pip install [-e] .``
+
+For a **development install**, use the optional extra `[dev]`.
 
 
 ### Pip-only install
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -28,36 +28,39 @@ can handle natively:
...
 ### Pip-only install
 If you are not going to use conda, then you must follow the following process (the reason
 that this is not a one-step process is because some of the packages inadvertantly require
 other packages to be installed before they can even be installed, let alone used, and ``pip``
 reads the ``setup.py`` file for each package before installing anything).
 
-First install the basic dependencies:
-
-``$ pip install numpy``
+First install `numpy`, as one of the dependencies requires it to be installed before
```
r-pascua, line 52: Though it might sound a little redundant, I think it would be clearer to say "... as one of the dependencies requires `numpy` to be installed ..."

r-pascua (approved):
Looks good to me, though there's one spot where the wording could be improved slightly. I checked the instructions using a fresh environment, and it worked (though I did have to install `git` first, so maybe we want to add a note about that?).

PR #132, opened 2021-02-25
Refactor utils

### LINKED ISSUE #125: Refactor `utils`, add filter cached property to `Simulator`

steven-murray:
Partially addresses #125

Converts `test_utils.py` to use pytest, does some refactoring of the `utils.py` module, and adds cached fringe-rate/delay filters to the `Simulator`.


r-pascua (approved):
This all looks great, thank you! I'm happy to approve as is, though I'd like to make a few comments regarding tests. 

- If we're trying to test as many lines of code as possible, then shouldn't we add some tests to ensure appropriate errors are raised if insufficient data is provided to `utils.rough_delay_filter` and `utils.rough_fringe_filter`?
- Do we want to add any checks for expected behavior when a pre-computed filter is passed to the `rough_x_filter` functions? i.e. asserting that the full signal is recovered (up to errors sourced from the FFTs) if we provide a filter that consists only of ones

In addition to the above, I'd personally prefer to have the bare `assert` statements be replaced with appropriate error raising (I recall reading an article explaining that using bare `assert` statements is unsafe in the sense that certain modes of running Python can cause `AssertionError`s to be ignored), but I won't demand that they be changed.

~~Regarding the failing tests: it looks like we've got an issue with `bda`; I'll look into this.~~
I've fixed the broken tests: there was a change to a parameter name in `bda.bda_tools.apply_bda` that was causing the CLI to break when applying BDA to the simulated data.

steven-murray:
@r-pascua thanks for the comments, which I completely agree with.

Did you also add the filter of ones? If not, I can try doing that.

r-pascua:
I did not add the filter of ones, but I can do that relatively quickly.

PR #133, opened 2021-02-26
Visibility docs

### LINKED ISSUE #58: Sphinx documentation

steven-murray:
Fixes #58 

Makes the API reference page a little nicer.


r-pascua (approved):
The API reference seems to be much more helpful now! I've noticed that there are a few classes that have boolean attributes that are used under the hood but may be useful to see at a glance in the API reference without needing to actually click on the thing (e.g. the `is_multiplicative` attribute of e.g. `DiffuseForeground`). I imagine it's not easy to do expose this from the doc end of things, so maybe we can update the code to make these properties with the `@property` decorator (assuming the overhead for that isn't annoying)? Either way, happy to approve this PR as-is; just wanted to mention this.

PR #139, opened 2021-05-02
Polarization capabilities for vis_cpu

philbull:
Adds experimental polarisation handling into `vis_cpu`. This supports polarised beams in the direct interpolation and pixel beam modes in `vis_cpu`, but not (yet) for `vis_gpu`. Only feed polarizations are used, i.e. `n` and `e` feeds; the user has to construct pseudo-Stokes visibilities manually. A pure Stokes I sky model is currently assumed.

This PR also adds a simple dipole-modulation polarisation model for `PolyBeam` that is inherited by `PerturbedPolyBeam`.

### DIFF
--- a/hera_sim/beams.py
+++ b/hera_sim/beams.py
@@ -145,8 +145,8 @@ class PolyBeam(AnalyticBeam):
         if self.polarized:
             interp_data[:,:,:,:] = modulate_with_dipole(az_array, beam_values)
         else:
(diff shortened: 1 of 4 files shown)


steven-murray:
@philbull this can be closed now, right?

philbull:
Yes, in deference to the equivalent PR against vis_cpu

PR #140, opened 2021-05-19
maint: remove vis_cpu function form package

steven-murray:
This simply removes the vis_cpu function, since it is now located at hera-team/vis_cpu. All functionality should remain the same.


steven-murray:
@hughbg this is now all passing. Let me know if you think I should change anything!

hughbg (approved):
All ok.

PR #141, opened 2021-05-19
update-docs

steven-murray:
Add in missing docstrings for v1.0.0.


hera_sim/simulate.py
```diff
@@ -61,39 +76,61 @@ def __init__(self, data=None, defaults_config=None, **kwargs):
...
+        """Unique simulation JDs."""
         return np.unique(self.data.time_array)
 
     def apply_defaults(self, config, refresh=True):
-        # TODO: docstring
-        """
-        """
+        """Apply a given set of defaults."""
         # actually apply the default settings
+        # TODO: SGM: why is this a method?
```
r-pascua, line 102: I suppose it doesn't need to be. I'm thinking that the motivation was that it would be convenient to allow a `Simulator` instance to control every aspect of the simulation, default parameters included.

hera_sim/simulate.py
```diff
@@ -61,39 +76,61 @@ def __init__(self, data=None, defaults_config=None, **kwargs):
...
+    def add(
+        self, component: Union[Type[SimulationComponent], str], **kwargs
+    ) -> Optional[UVData]:
+        """Add a particular simulation component to the simulated data.
+
+        Parameters
+        ----------
+        component
+            Either a string name (or alias) of a component model to add,
+            or its actual class.
```
r-pascua, line 114: It can also be an instance of the class.

hera_sim/simulate.py
```diff
@@ -61,39 +76,61 @@ def __init__(self, data=None, defaults_config=None, **kwargs):
...
+            or you just want a reference to this particular component without being
+            mixed with the rest of the simulation.
+
+        Returns
+        -------
+        Optional[UVData]
+            If ``ret_vis=True``, the data simulated from this component.
+        """
+
+        # TODO: why aren't these just in the signature itself?
```
r-pascua, line 133: They (as well as some of the other optional parameters) are moved to the signature in the branch where I tidy stuff up in the `Simulator` code. I think at the time of writing I didn't know exactly what optional arguments I wanted/needed, so just left the ``kwargs`` as a catch-all. This shouldn't be an issue once I've finished tidying up the code.

r-pascua (commented):
This is looking really great, thank you for doing this! I think the only bit of documentation that hasn't been touched in this PR that needs to be updated is the actual module-level documentation. We might be able to just copypasta most of the stuff from the main branch of the package instead of writing new module-level descriptions.

Thanks again for your work! I'll try to review ASAP once this is finished. You can leave the `Simulator` documentation as work for me to do, since some of the methods either no longer exist or have had their signatures have changed in my dev branch.

hera_sim/defaults.py
```diff
@@ -31,30 +31,30 @@ class Defaults:
...
-    hera_sim.defaults(parameter),
+        hera_sim.defaults(parameter),
 
     where `parameter` is a string with the name of the parameter as
     listed in the configuration file. To view the entire set of default
-    parameters, use:
+    parameters, use::
 
-    hera_sim.defaults()
+        hera_sim.defaults()
```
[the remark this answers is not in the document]
r-pascua: To follow-up on this, there are some examples (I think in the `sigchain` module?) where only a single colon precedes the actual example, rather than the sets of two colons used here.

steven-murray:
Thanks @r-pascua for the review so far -- super useful. I'll try to get to the comments tomorrow 🤞🏻 .

hera_sim/defaults.py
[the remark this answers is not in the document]
steven-murray: Yeah, RST requires a double-colon before code blocks (and they should be indented as they are here). I'll go find other examples to fix them too :-)

hera_sim/sigchain.py
```diff
@@ -86,16 +109,44 @@ class Reflections(Gain, is_multiplicative=True):
...
         self, amp=None, dly=None, phs=None, conj=False, amp_jitter=0, dly_jitter=0
     ):
-        # TODO: docstring
-        """
+        """Produce multiplicative reflection gains.
+
+        Parameters
+        ----------
+        amp : float, optional
+            Mean Aamplitude of the reflection gains.
```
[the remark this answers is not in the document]
steven-murray: Done, thanks

hera_sim/rfi.py
```diff
@@ -105,7 +168,7 @@ def __call__(self, lsts, freqs, **kwargs):
         for station in stations:
             if not isinstance(station, RfiStation):
                 if len(station) != 5:
-                    raise ValueError(
+                    raise TypeError(
```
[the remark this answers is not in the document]
steven-murray: Got it. Makes sense.

hera_sim/simulate.py
```diff
@@ -34,18 +36,31 @@ def new_func(*args, **kwargs):
...
+        ----------
+        data
+            Any input data. If a path, will load a set of
+            visibilities using the ``read()`` method of
+            :class:`UVData`. Can be a :class:`UVData` object
+            to initialize directly. If not set, initialize an
+            empty :class:`UVData` object.
+        defaults_config
+            If given, either a path pointing to a defaults configuration
+            file, or a dictionary of configuration parameters (see :class:`defaults.Defaults`).
```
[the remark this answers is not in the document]
steven-murray: Cool. Fixed.

hera_sim/simulate.py
steven-murray: Good to know. Fixed.

hera_sim/simulate.py
steven-murray: OK cool. removing my comment :-)

hera_sim/sigchain.py
```diff
@@ -123,8 +174,28 @@ def __call__(self, freqs, ants, **kwargs):
...
+        Parameters
+        ----------
+        freqs : array_like of float
+            Frequencies, units GHz.
+        amp : array_like of float
+            Either a scalar amplitude, or 1D with size Nfreqs, or 2D
+            with shape (Ntimes, Nfreqs).
+        dly : [type]
+            Either a scalar delay, or 1D with size Nfreqs, or 2D
+            with shape (Ntimes, Nfreqs).
```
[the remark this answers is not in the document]
steven-murray: Yup, I went the more general route.

steven-murray:
@r-pascua -- I've made quite a few updates here (all docs stuff still). Most importantly, I've turned on a lot of flake8 checks for documentation, which should keep us up-to-date going forward. I've also address all the errors that came up because of those checks EXCEPT those to do with the `Simulator` class (I'll wait to pull in your changes here before fixing those up). 

Tests are failing because I'm waiting on #140 to be merged, but otherwise the changes could be reviewed. I've checked most of the docs by building and viewing them, and suggest that as the easiest way to check them. 

Sorry about the 41 files changed!

hera_sim/io.py
```diff
@@ -29,10 +28,40 @@ def empty_uvdata(
...
+    start_freq : float, optional
+        Lowest frequency channel, by default None
+    channel_width : float, optional
+        Channel width, by default None
+    n_freq : [type], optional
+        Alias for ``Nfreqs``
+    n_times : [type], optional
+        Alias for ``Ntimes``.
+    antennas : [type], optional
+        Alias for array_layout?
```
r-pascua, line 52: This is indeed an alias for `array_layout` and is kept purely in the interest of backwards compatibility.

hera_sim/io.py
```diff
@@ -29,10 +28,40 @@ def empty_uvdata(
...
+    """Create an empty UVData object with given specifications.
+
+    Parameters
+    ----------
+    Ntimes : int, optional
+        NUmber of unique times in the data object.
+    start_time : float, optional
+        Starting time (Julian date) by default 2456658.5
+    array_layout : [type], optional
+        Specify an array layout.
```
r-pascua, line 40: The typing for this is `Dict[int, np.ndarray]`.

r-pascua (commented):
This is looking really great, thank you! I just have a few notes I want to bring up as discussion items, either to be addressed here or later.

- Is it possible to make readthedocs ignore some classes? (in particular the abstract classes like `Foreground`, etc.) I'm not sure if it's useful having those classes show up in the API reference.
- When referencing another method/function in a docstring, (e.g. with `:func:~utils.rough_delay_filter`), are the docs supposed to contain a link to the actual thing referenced? If so, then this feature doesn't seem to be working correctly (see e.g. docs for `DiffuseForeground`).
- It looks like the `__call__` method docs aren't making it to the readthedocs for any class. From a documentation standpoint, I think the thing to do here is to update `SimulationComponent._update_call_docstring` to instead modify the class docstring, adding the information from the `__call__` docstring. I don't know if this will actually make all of the parameters show up in the API reference, but it's worth a shot (right now none of the classes have the positional arguments show up in the docs).
- The `defaults` module reference is virtually empty, containing only the module docstring.
- Is it possible to flag certain methods to be documented in a separate section from the rest of the methods? In particular, I'd like the legacy `Simulator.add_x` methods to live in their own box.
- The "Other Parameters" section is getting formatted strangely. I think that readthedocs is expecting the name of a parameter followed by a description, so it's just boldfacing the sections and not doing any of the typesetting you'd expect it to do (see `io.empty_uvdata` reference for an example).
- We need to update some of the higher-level docs (e.g. contributing and installation).

I think it would be good for us to briefly discuss these points before I hit that shiny "approve" button. I don't necessarily think anything needs to be changed, but I do see that there are a few things that still need to be finished. You've already done a lot of work on this end, so we can always opt to get these changes in now and then have one more final "close inspection"-type PR before the version 1 release.

hera_sim/io.py
steven-murray: Thanks, added that in.

hera_sim/io.py
steven-murray: Gotcha.

steven-murray:
> Is it possible to make readthedocs ignore some classes? (in particular the abstract classes like Foreground, etc.) I'm not sure if it's useful having those classes show up in the API reference.
   
Just to be clear, it's not RTD that's processing the docs, but Sphinx. There are a couple of ways to do this, but none of them are particularly pretty:

1. We could add an empty docstring to the classes, but then we'd also have to add a `# noqa` to stop flake8 checking on that line, AND we don't get the internal documentation. 
2. We could add a leading underscore to the abstract classes. This to me seems like the right way to do, since these classes really are "private" anyway. However, this will break the getting of components by their names as it currently stands. We'll have to strip the leading underscore in the plugin dict keys.
3. We can manually add each desired ignored class to the `:undoc-members:` list in the RST. But I think this is going to be really hard due to the way I've set up the autosummary. 
4. We could leave the base classes there, and add a little more docs to each to provide an overall summary of that kind of component. 

> When referencing another method/function in a docstring, (e.g. with :func:~utils.rough_delay_filter), are the docs supposed to contain a link to the actual thing referenced? If so, then this feature doesn't seem to be working correctly (see e.g. docs for DiffuseForeground).

Yes, they are supposed to link... I'll have to look into that. 
  
>  It looks like the __call__ method docs aren't making it to the readthedocs for any class. From a documentation standpoint, I think the thing to do here is to update SimulationComponent._update_call_docstring to instead modify the class docstring, adding the information from the __call__ docstring. I don't know if this will actually make all of the parameters show up in the API reference, but it's worth a shot (right now none of the classes have the positional arguments show up in the docs).
  
I've fixed the class template in sphinx to include the `__call__` docstring, and at the same time removed the `__init__` (since we document the class in the top-level docstring now). Let me know if that looks right to you.

 > The defaults module reference is virtually empty, containing only the module docstring.
    
Hmmm, you're right. No idea why this is being treated differently. I'll look into it. 

> Is it possible to flag certain methods to be documented in a separate section from the rest of the methods? In particular, I'd like the legacy Simulator.add_x methods to live in their own box.
    
Yes, though it's not overwhelmingly easy. I'll give it a go, but no promises. We could also consider not keeping them in the online docs since they are legacy.

> The "Other Parameters" section is getting formatted strangely. I think that readthedocs is expecting the name of a parameter followed by a description, so it's just boldfacing the sections and not doing any of the typesetting you'd expect it to do (see io.empty_uvdata reference for an example).
    
Yup, fair enough. I'll have a bit of a look around at what you're supposed to do here. 

> We need to update some of the higher-level docs (e.g. contributing and installation).

OK, I'll try do that in this PR.

steven-murray:
1. I chose to go the route of adding more docs to the base component classes, to make them somewhat useful -- this is done using the `component` decorator (so the added docs are fairly generic). I can revert this and remove them altogether if you think that's better.
2. I think there were just errors in the `utils.py` that caused the links not to work. They should be working now.
3. I added the __call__ docstrings.
4. I cannot figure out why the `defaults` module is not getting its members documented. I don't know where to look to figure it out. I'll create an issue for it, and let's just get on with the PR.
5. Check the `Simulate` class docs out now: there's a new section for Deprecated methods.
6. I've fixed all the Other Parameters sections according to the LSST style guide.
7. I've streamlined some of the top-level docs, removing references to aipy and making the install process simpler.

r-pascua:
This is looking very good. Thank you for all your hard work! Happy to merge this whenever, though there are a few things I'd like to point out.

- Inter-module references still don't seem to work (e.g. the references to the filtering functions in the `DiffuseForeground` docs), but intra-module references seem to be fine.
- It looks like I messed up the mathmode docs in `sigchain.vary_gains_in_time`. Not important to fix now, but just want to take note of it somewhere.
- There are some module-level docs that need to be updated (`foregrounds`, `noise`, `rfi`, and `sigchain` need to be updated).
- The installation section on the RTD still says `aipy` is a requirement (I think we've moved past `aipy`, no?). This may be a bug with Sphinx using some cached stuff for generating the docs (it's done this to me before), but I figured I'd note it anyway.

I'm going to take one last pass over #143 before merging that into `v1.0.0`, then I think we're good to go.

steven-murray:
> Inter-module references still don't seem to work (e.g. the references to the filtering functions in the DiffuseForeground docs), but intra-module references seem to be fine.
   
I fixed this -- turns out reference have to be fully qualified (i.e. needed to make it either `~.utils.rough_delay_filter` (note preceding dot) or `~hera_sim.utils.rough_delay_filter`. I went through and fixed all the cases I could find. At the same time, fixed all the references to other codes, and added intersphinx mappings, so now pyuvdata references actually link to their docs!

> It looks like I messed up the mathmode docs in sigchain.vary_gains_in_time. Not important to fix now, but just want to take note of it somewhere.

Fixed it.   

> There are some module-level docs that need to be updated (foregrounds, noise, rfi, and sigchain need to be updated).

I updated all of these to be not awful. But they're still very basic. Happy if you want to update them further, but also happy if not.
   
> The installation section on the RTD still says aipy is a requirement (I think we've moved past aipy, no?). This may be a bug with Sphinx using some cached stuff for generating the docs (it's done this to me before), but I figured I'd note it anyway.

This was pointing to an old file which I hadn't realized. I updated the docs index to point directly at the top-level README (which I also converted to RST and cleaned up a little, and added a docs badge), which has an installation section in it. 

I am now happy for this to be merged, once we get the green light.

r-pascua (approved):
This is amazing! The cross-referencing works fantastically now. I've got just a few brief notes:
1. There's a file `.vscode/settings.json` that is being tracked that I don't think should be tracked (it just has your local python path in it).
2. This RTD build didn't include the `README`, even though the config says to do so. Not sure if this is another Sphinx caching error or a genuine bug (though I'm leaning toward the former).
3. Some modules are having trouble with their docs. We can punt on getting this fixed, but I figure I should take note: `defaults` still doesn't want to document anything but the module docstring; `cli_utils` and `components` do not show up in the API reference, and I don't really understand why.

Thank you for figuring all of this out and getting everything formatted so nicely! I'm happy with getting this merged into `v1.0.0` whenever.

steven-murray:
OK, deleted the vscode/ file (woops). I had a wrong link to the README, so fixed that. And I added `cli_utils` and `components` to the docs (I had just forgotten them). However, I still can't figure out the defaults thing. I'll make an issue.

PR #143, opened 2021-06-03
Finalize simulator

r-pascua:
This PR tidies up and prunes much of the code in the `Simulator` class, completes the documentation for the class, and brings its coverage up to 95%. There are some other changes that have been made along the way; however, I can't recall all of the modules affected off the top of my head. I think there are still a few bits of documentation I have to finish fleshing out, but these will be done soon. Use of cached fringe/delay filters has been implemented, and it works for speeding up simulations for large arrays with a decent degree of redundancy (though more precise benchmarking has not been done).

There are a few outstanding items, though:
- Using a version of `pyuvdata` that is at least 2.0 causes some issues regarding redundancy of the simulated results. I've traced this back to an error related to preserving the input array layout, and I've messaged the `pyuvdata` folks about this to start figuring out how to resolve the issue. For now, the failing tests have been marked with an `xfail`.
- The `Simulator._iteratively_apply` code is much cleaner now, though I'm not sure how understandable it is to other folks. I also think that there might be a smarter way to deal with the whole returning vs adding business than how I've settled on dealing with it.
- Cached filters are implemented by calculating a delay and fringe filter for each redundant group, and the user must call the `Simulator.calculate_filters` method with the desired parameters (just once, *prior to using the `Simulator.add` method*) in order for the filters to be cached (otherwise they are always generated on the fly and return to the ether after the effect is simulated). This isn't optimal, since the fringe filters only depend on EW baseline length and the delay filters only depend on total baseline length, but it works and is a reasonably good approximation of the "correct" way to do this.
- There may be some merge conflicts I need to work out when merging in the current version of `v1.0.0`... I'll get to that later, but certainly before this PR is to be merged.
- There's a check on `yaml` files that claims the constructor for the `!antpos` tag can't be found. This constructor exists, and the tag is in three configuration files, but it only raises the issue for one of the files. No clue what's going on here, especially since we don't have any configuration files for how the `yaml` files are verified.

I think after this PR, as well as #141 and #90, are merged in, then the version 1 release will only be blocked by the changelog needing updating. I'm debating whether it would be worthwhile to make a tutorial notebook or something that showcases differences between version 1 and pre-version 1, but I suppose we can discuss this later.

Finally, I ran into some issues regarding basic polarization support (in anticipation of future updates that add full polarization support). I ended up sorting these out, but I think this is something we'll need to think about carefully in the future.


hera_sim/simulate.py
```diff
@@ -31,226 +39,386 @@ def new_func(*args, **kwargs):
...
-
+    def __init__(
+        self,
+        *,
+        data: Optional[Union[str, UVData]] = None,
+        defaults_config: Optional[Union[str, Dict]] = None,
+        redundancy_tol: float = 1.0,
+        **kwargs,
+    ):
+        """Simulate visibilities and instrumental effects for an entire array.
```
steven-murray, line 55: In the `update-docs` branch I decided to go with putting the class docstring at the class level instead of in `__init__`, and turned off checks for having docs in `__init__`. I had a good look around the web to check what was best practice and tbh no-one really knows. But the numpydoc standard seems to indicate they should be put in the class docs.

hera_sim/simulate.py
```diff
@@ -31,226 +39,386 @@ def new_func(*args, **kwargs):
...
+            dictionary for setting default simulation parameters. See tutorial
+            on setting defaults for further information.
+        redundancy_tol
+            Position tolerance for finding redundant groups, in meters. Default is
+            1 meter.
+        kwargs
+            Parameters to use for initializing UVData object if none is provided.
+            If ``data`` is a file path, then these parameters are used when reading
+            the file. Otherwise, the parameters are used in creating a ``UVData``
+            object using ``io.empty_uvdata``.
```
steven-murray, line 73: use `:func:~io.empty_uvdata`

hera_sim/simulate.py
```diff
@@ -31,226 +39,386 @@ def new_func(*args, **kwargs):
...
+
+        Attributes
+        ----------
+        data: ``pyuvdata.UVData``
```
steven-murray, line 77: colons need to have a space before them

hera_sim/simulate.py
```diff
@@ -31,226 +39,386 @@ def new_func(*args, **kwargs):
...
+        """Array of polarization strings."""
+        return self.data.get_pols()
+
     def apply_defaults(self, config, refresh=True):
-        # TODO: docstring
         """
+        Apply the provided default configuration.
+
+        Equivalent to calling ``hera_sim.defaults`` with the same parameters.
+        See ``hera_sim.defaults.set`` documentation for further details.
```
steven-murray, line 161: use `:meth:`

hera_sim/simulate.py
```diff
@@ -31,226 +39,386 @@ def new_func(*args, **kwargs):
...
+        vis_filter: Optional[Sequence] = None,
+        name: Optional[str] = None,
+        **kwargs,
+    ):
+        """
+        Simulate an effect then apply and/or return the result.
+
+        Parameters
+        ----------
+        component: str or SimulationComponent subclass
```
steven-murray, line 204: again with the spaces before colons, and use `:class:` for the sim component.

hera_sim/simulate.py
```diff
@@ -31,226 +39,386 @@ def new_func(*args, **kwargs):
...
         """
+        Retrieve an effect that was previously simulated.
+
+        Parameters
+        ----------
+        component
+            Effect that is to be retrieved. See .. meth:: add for more details.
+        key
+            Key for retrieving simulated effect. Possible choices are as follows:
+                An integer may specify either a single antenna (for per-antenna
```
steven-murray, line 285: I think Sphinx is going to complain about the extra indent here.

hera_sim/simulate.py
```diff
@@ -31,226 +39,386 @@ def new_func(*args, **kwargs):
...
+        component: Component,
+        key: Optional[Union[int, str, AntPair, AntPairPol]] = None,
+    ) -> Union[np.ndarray, Dict[int, np.ndarray]]:
         """
+        Retrieve an effect that was previously simulated.
+
+        Parameters
+        ----------
+        component
+            Effect that is to be retrieved. See .. meth:: add for more details.
```
steven-murray, line 282: use `:meth:add`

hera_sim/simulate.py
```diff
@@ -367,118 +537,292 @@ def chunk_sim_and_save(
     # TODO: write a deprecated wrapper function
     def add_eor(self, model, **kwargs):
         """
-        Add an EoR-like model to the visibilities. See :meth:`add` for
+        Add an EoR-like model to the visibilities. See .. meth:: `add` for
```
steven-murray, line 540: I'm pretty sure the original is correct here? The `.. meth::` form is for its own line.

setup.cfg
```diff
@@ -15,6 +15,7 @@ project_urls =
     Documentation = https://hera_sim.readthedocs.org
 # Change if running only on Windows, Mac or Linux (comma-separated)
 platforms = any
+version = 1.0.0
```
steven-murray, line 18: This should not be necessary

steven-murray:
Merging v1.0.0 back into here should fix the RTD issue, I think. As for the YAML issue.... I'm not sure 🤷🏻

hera_sim/simulate.py
r-pascua: I moved the `__init__` docstring to the class docstring for the `Simulator`. Not sure if we should have anything in the `__init__` docstring, so it's currently empty.

hera_sim/simulate.py
r-pascua: Addressed this and all other occurrences of this issue.

hera_sim/simulate.py
r-pascua: Done.

hera_sim/simulate.py
r-pascua: Done, though I'm not sure if the updated version is exactly right? I put `:meth:` before both references in this docstring.

hera_sim/simulate.py
r-pascua: Done.

hera_sim/simulate.py
r-pascua: Sphinx didn't have any issues with it, and it looks fine on the RTD! Though looking through the RTD made me realize there are some other formatting issues I need to fix.

hera_sim/simulate.py
r-pascua: That's right. I've reverted this change.

hera_sim/simulate.py
r-pascua: Done.

setup.cfg
r-pascua: I've removed it.

r-pascua:
OK, I think I've got everything worked out and filled out the last of the docstrings. Some (but not all) internal references are still wonky on the RTD, but I'm not going to stress over that in this PR.

The YAML issue was that `pre-commit` was actually trying to compile the files, but was having an issue. Strangely this was only happening for one of three configs, but each config has that same constructor. :shrug: Either way, I updated the `pre-commit` config so that YAML files are only checked for syntax and are *not* compiled, which seems to have fixed the problem.

I'm going to wait a little bit to merge this. I want to take one more look over everything with fresh eyes to make sure I didn't miss anything. I should have it merged within the next 24h though.

steven-murray:
@r-pascua having a quick look at the `Simulator` docs, I think it overall looks good. A couple of things I noticed though:

* You reference private functions (eg. `_seed_rng`) in the docstrings for public functions (i.e. `add`). However, by default the private functions are not output in sphinx docs. We can forcibly output them, but I think it's better practice to either not reference them, or to make them public.
* Weirdly, the signature generated from the Types by sphinx is not linking to classes within hera_sim. This might be resolved on `update-docs`, so I wouldn't worry too much, just pointing out in case.

r-pascua:
OK, instead of referencing private methods in public docstrings, I've referred the reader to the tutorial for using the `Simulator`&mdash;I'll just need to remember to add a section on advanced usage when fleshing out the `Simulator` tutorial. I chose this because I think the methods I referenced (`_seed_rng` and `_apply_filter`) really aren't meant to be interfaced with by the user, but they are well-documented with the intention that folks who are worried about how the guts of the class work can read about it. I'll worry (or won't, depending on how it turns out) about the signature thing in #141 after this is merged.

PR #145, opened 2021-06-16
Update the changelog

r-pascua:


steven-murray (approved):
This looks awesome! Thanks for the extremely detailed changelog update, which will make it much easier for others to update to v1.0.0. 

A side-note: I'm not sure that using the `:func:...` notation in the changelog is compatible with releasing on pypi. It's definitely nice for the html docs, so we'll have to cross that bridge when we get there.

r-pascua:
Ah, that's a good catch. If it ends up being an issue for the pypi release, then it should be simple enough to change (and I'd be happy to do it).

PR #146, opened 2021-06-17
Simulator tutorial

r-pascua:
Updates the tutorial for the `Simulator`, showcasing basic use and advanced use. For some reason the history strings are saying that it's using `v0.0.0` of `hera_sim`... not sure what's up with that.


steven-murray (approved):
All looks good to me!

steven-murray:
Ugh. Looks like we need another test or two.

r-pascua:
Great! I'll get the coverage to a level where it won't complain, then go ahead and start the grand merging.

PR #147, opened 2021-06-17
Enable new vis_cpu coordinate system treatment

philbull:
This PR enables the new coordinate system handling features for VisCPU associated with this PR:
https://github.com/HERA-Team/vis_cpu/pull/11

This includes a method that allows more accurate handling of equatorial coordinates, and has more explicit documentation on what coordinates are assumed. This also enables basic polarization support.

Target: v1.0.0.


setup.cfg
```diff
@@ -43,7 +43,7 @@ install_requires =
     pyuvdata>=2.0<3.0
     astropy_healpix
     astropy
-    vis_cpu>=0.1.1
+    vis_cpu
```
steven-murray, line 46: I would leave this with the >=, since we really do need that version or greater.

setup.cfg
```diff
@@ -70,10 +70,12 @@ tests =
     pytest-cov>=2.5.1
     pre-commit
     matplotlib>=3.4.2
+    IPython
```
steven-murray, line 73: I couldn't see what IPython was being used for?

setup.cfg
```diff
@@ -70,10 +70,12 @@ tests =
...
     uvtools @ git+git://github.com/HERA-Team/uvtools.git
     hera_cal @ git+git://github.com/hera-team/hera_cal
     healvis @ git+git://github.com/rasg-affiliates/healvis
     bda @ git+git://github.com/HERA-Team/baseline_dependent_averaging.git
+    vis_cpu @ git+git://github.com/HERA-Team/vis_cpu.git
```
steven-murray, line 78: We're trying to release hera_sim as a PyPI package, and as such it can't have git-url requirements like this (I've made issues on all the other repos above this to this effect as well). We should just release a new vis_cpu version so that this branch works with the PyPI version of vis_cpu.

steven-murray (commented):
Thanks @philbull -- just a few fixes/clarifications to be made. I can release a new vis_cpu if that helps as well.

hera_sim/tests/test_vis.py
```diff
@@ -239,6 +239,10 @@ def test_autocorr_flat_beam(uvdata, simulator):
     )
     v = sim.simulate()
 
+    # Account for factor of 2 between Stokes I and 'xx' pol for vis_cpu
+    if simulator == VisCPU:
+        v *= 2.0
+
```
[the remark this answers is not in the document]
philbull: This is the fix I guess -- I switched vis_cpu to applying a factor of 1/2 to the input source fluxes (assumed to be specified as Stokes I), as explained here: https://github.com/HERA-Team/vis_cpu/blob/main/src/vis_cpu/vis_cpu.py#L92
The rest of vis_cpu naturally deals with linear polarisation channels (rather than Stokes), so the output in this case is for the 'xx' polarisation and so needs to be multiplied by a factor of 2 again to get Stokes I.

This change makes the output consistent with pyuvsim, which you can see from this unit test in the vis_cpu package: https://github.com/HERA-Team/vis_cpu/blob/main/tests/test_compare_pyuvsim.py#L143

But in _this_ unit test (in hera_sim/tests/test_vis.py), the extra factor of 2 is needed, as I assume the other simulators are outputting Stokes I.

So, I'm not sure what the best course of action is here. I think the new vis_cpu behaviour makes sense, as it is now a polarised visibility simulator, and it matches pyuvsim.

setup.cfg
philbull: Thanks, I've actually bumped this to >=0.2.0, since this PR uses recent changes in vis_cpu

setup.cfg
philbull: Thanks, this was added in this PR just to get the tests to work properly. Now the vis_cpu version has been bumped I've removed it.

hera_sim/tests/test_vis.py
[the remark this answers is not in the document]
steven-murray: Hmm, OK if vis_cpu is consistent with pyuvsim, it probably means we instead need to update the HealVis wrapper to return results consistent with pyuvsim. This has been a persistent problem, and has not been helped by my own very vague understanding of polarization etc.

steven-murray (approved):
Happy for this to go in -- but if you feel like taking a look at the HealVis wrapper, it might be worth fixing that up and then changing these tests back so that no *2 is required. Otherwise, we can create an issue for it and resolve it independently.

hera_sim/tests/test_vis.py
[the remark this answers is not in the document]
philbull: Thanks, I can look into this if needed, but perhaps it's more sensible for the HealVis maintainer to do it (not sure who at the moment?)

setup.cfg
philbull: This was mistakenly included. There was an issue with vis_cpu importing IPython, but not actually including it properly in its install dependencies. I have now made it optional in this PR: https://github.com/HERA-Team/vis_cpu/pull/13

philbull:
Thanks Steven. I think it's best to merge this in now if you're happy with it, and then look at HealVis separately, as in my experience anything to do with polarisation is a can of worms...

Issue #85, opened 2020-02-15
Investigate/Fix `foregrounds.pntsrc_foreground`


r-pascua:
While fixing the tutorial notebooks, I noticed that the `pntsrc_foreground` function doesn't seem to be working exactly as it should. In particular, it looks like the sources aren't actually being convolved with the beam. This is most easily seen in the diff between the old and new `hera_sim_tour` notebook output:
![pntsrc_foreground_issue](https://user-images.githubusercontent.com/29933285/74595719-d65b9a80-4ff9-11ea-88b0-9bc03d688c8e.png)

r-pascua:
Not really sure why this was breaking in the updated version of the notebook. Here's an example plot using the `pntsrc_foreground` function from v1. It looks like the beam convolution is actually happening here, so I think the problem is resolved in v1? Either way, the guts of the function could be updated to be... less mysterious.
![pntsrc_fg](https://user-images.githubusercontent.com/29933285/81993382-5f36a200-95fa-11ea-8dad-fb09473fb6f4.png)

r-pascua:
As I'm going through rewriting the tests, I'm doing some sanity checks on the original set of tests. I noticed that there was an orientation check for the diffuse foreground simulator, but not for the point source simulator, so I mocked up some point-source foregrounds and decided to look at the fringe-rate transform of some baselines. Long-story short, it looks like something isn't being done right, even in the context of a quick-and-dirty simulator. My understanding of how visibilities behave in fringe-rate space is still a little foggy, but I think the right way to think about it is that the fringe-rate transform tells you about how the fringe frequency evolves as the Earth rotates. So baselines with an eastward EW projection should have power at positive fringe rates, whereas those with a westward EW projection should have power at negative fringe rates, and purely NS baselines should have power at zero fringe-rate. (I think this should follow from an argument considering a single point source locked to the sky.) At the very least, I know that NS baselines should only have power at zero fringe-rate, but that's certainly *not* the case with the point source foreground simulator. Check it out:
![bad_pntsrc_sim](https://user-images.githubusercontent.com/29933285/88888803-d809cb00-d1f3-11ea-9633-50fb80c53fb5.png)

Strangely enough, the fringe-rate transform for a 100-meter EW baseline looks basically the same (didn't happen to save that figure, but this one should be more than enough to show there's a problem). I should note that the above was made with the following parameters:

- 1000 integrations spanning the course of a day
- 1024 frequency channels from 100-200 MHz
[...]

PR #155, opened 2021-06-24
Fix point source foreground sim

r-pascua:
This is meant to address #85. This PR just involves a minor bugfix in `foregrounds` as well as in the `Simulator` tutorial. I'm not sure if this necessarily closes the issue, since it looks like some of the simulated point sources actually get brighter with frequency (which doesn't seem right), but now the simulated visibilities look much better. (See the `Simulator` tutorial, in the "Naming Components" section, for an example.)


steven-murray (approved):
Looks good to me :-) Thanks @r-pascua

steven-murray:
Ah, I should have said: update the CHANGELOG!

Issue #152, opened 2021-06-24
Add better dev install instructions to README


steven-murray:
Especially mentioning the clone depth because the repo is quite large.

Make a new section for the dev install.

piyanatk:
Completed in #156

PR #156, opened 2021-06-29
Add developer install documentation and temporary require `pyuvdata<2.2.0`

piyanatk:
For #152.


steven-murray (approved):
Thanks @piyanatk !!

steven-murray:
@r-pascua any idea why these tests are failing? Only one file (the README) was changed, so it makes no sense to me. Was there an updated dependency?

r-pascua:
It looks like there was maybe an update to `pyuvdata` where the `UVData.check` method requires that the `phase_type` attribute be set. This is not currently set by the `io.empty_uvdata` function, so I imagine that's what's ultimately causing the error. I believe this can be fixed by inserting `uvd.set_drift()` in line 103 of the `io` module. I can do this, or someone else can.

r-pascua:
Fixing this on my PR so you don't need to do it here.

r-pascua:
Sorry for the slew of comments, but I figured it out. This is a problem with `bda`, so I'll raise an issue there. For now, I'm going to hit the `test_cli` test with an xfail and add the needed fix to the `io.empty_uvdata` function in my PR (not the best practice, but I thought it would be an easy fix and got ahead of myself). The core of the problem is that version 2.2 of `pyuvdata` changed how phasing works, and the `bda.apply_bda` function doesn't copy over any of the phasing information. This version also made the `phase_type` attribute a required parameter, so now the `check` method is saying the `UVData` object isn't configured correctly in the copy that has had BDA applied to it. This is a little annoying, but whatever. It should get fixed eventually.

piyanatk:
I updated the branch with @r-pascua update but it does not pass codecov. Any idea?

steven-murray:
@piyanatk it's because we have an extra branch of code that no longer fires in tests. I think probably what we should do is just put a `# pragma: nocover` on the pre-2.2.0 branch in `io`, since we're not testing against older versions of pyuvdata.

setup.cfg
```diff
@@ -40,7 +40,7 @@ install_requires =
     scipy
     cached_property
     pyuvsim>=1.1.2
-    pyuvdata>=2.0<3.0
+    pyuvdata>=2.0<=2.2.0
```
r-pascua, line 43: This should be `<2.2.0`, since the problem is from `2.2.0` onward.

r-pascua: Once you make this change, you should also remove the `@pytest.mark.xfail` from the `test_cli` function in the `tests/test_simulate_cli.py` script.

piyanatk:
It looks like deleting `xfail` decorator has failed the test.

setup.cfg
piyanatk: Changed to `<2.2.0`

r-pascua:
Ah, in addition to modifying the `setup.cfg` file, you need to update the `ci/tests.yaml` file, since that's what is used to set up CI. Just make the same edit and I think it should work. If the test fails again, then check the environment setup in the test details for which version of `pyuvdata` is being installed.

piyanatk:
Okay. All tests passed now. But `pyuvdata=2.2.0` is still being install first in environment setup in the test because of these lines. 
https://github.com/HERA-Team/hera_sim/blob/bb942c6cb300dadf2c9d9a671177f303f1ca7830/.github/workflows/test_suite.yaml#L38-L43
It is later downgrade to `2.1.4`, so everything works fine. Should we leave this for now.

steven-murray:
Thanks @piyanatk  --  I think we can leave that. We could also change that line to have `pyuvdata<2.2.0`, but really I hope this doesn't last long before we can take the pin away. Let's leave it and merge.

piyanatk:
Okay. Cool. I will merge now.

plaplant:
I don't mean to butt in here, but what features of pyuvdata v2.2.0 specifically are breaking tests? I saw that @r-pascua had mentioned the `phase_type` issue https://github.com/HERA-Team/baseline_dependent_averaging/issues/5, but was curious if there was anything else.

aewallwi:
Potentially related, the change in the `pyuvdata` API have broken some `hera_spec` code. 

https://github.com/HERA-Team/hera_pspec/issues/344

The issue also involves `hera_sim` changes which are currently being covered but we don't have anyone addressing the `pyuvdata` changes yet.

Maybe someone here can help with this?

r-pascua:
@plaplant the `bda` interface in the CLI (which boils down to the `phase_type` issue) is the only thing that is breaking from pyuvdata v2.2.0. That said, whenever the new array shapes are made the default behavior, (I expect that) things will break and it will require some updates on my end to fix them. To that end, it would be nice to have some warning of when that change is expected to be made so I know when to prioritize figuring out the fix.

plaplant:
@r-pascua thanks, I should be able to fix that `phase_type` issue relatively easily. We're not planning on making the new array shapes change until pyuvdata v3.0.0, which we don't have a specific date for yet, but shouldn't happen for a little while. I can give you a heads up when we have a firmer timeline.

PR #157, opened 2021-07-01
Fix Simulator LST array

r-pascua:
The previous implementation had a bug that caused the `lsts` attribute to be calculated incorrectly if a blt-order other than time-baseline was used. The new implementation uses the sorting on the time array in order to retrieve the unique LSTs while respecting phase wraps. I've also added a test that ensures the unique LSTs are recovered correctly when a non-default blt-order is used.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,16 +5,11 @@ Changelog
 v1.0.2 [2021.07.01]
 ===================
 
-Added
------
-
 Fixed
 -----
 - Bug in retrieval of unique LSTs by :class:`~.Simulator` when a blt-order other than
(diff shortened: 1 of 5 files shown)


CHANGELOG.rst
```diff
@@ -2,6 +2,20 @@
 Changelog
 =========
 
+v1.0.2 [2021.07.01]
+===================
+
+Added
```
steven-murray, line 8: Remove headings that don't have content.

steven-murray (approved):
Thanks @r-pascua, this looks good to me -- just remove the unnecessary headings from the CHANGELOG. Once it's merged, either let me know, or do the following on your locall system:

```
git pull
git tag v1.0.2
git push --tags
```

PR #160, opened 2021-07-13
fix conversions.

### LINKED ISSUE #159: `vis_cpu` incorrectly calls `uvbeam.interp`
As we can see from

https://github.com/RadioAstronomySoftwareGroup/pyuvdata/blob/4c66208245db29be8b9920434d6d48d23b29e367/pyuvdata/uvbeam/uvbeam.py#L1579

`uvbeam.interp` takes `az_array` and `za_array` as its first two args but `freq_array` is now the sixth arg. 

Unfortunately, `conversions.py` calls `uvbeam.interp` assuming that `freq_array=freqs` is the third arg

https://github.com/HERA-Team/hera_sim/blob/144f8ec8033e45ace1a30e3f1f45c8a2489512b3/hera_sim/visibilities/conversions.py#L36

which assigns `freqs` to `az_za_grid` in `uvbeam.interp` and causes all kinds nonsense.

This line should be changed to name the args.

` efield_beam = uvbeam.interp(az_array=az, za_array=za, freq_array=freqs, **kwargs)[0]`

aewallwi:
closes #159

### DIFF
--- a/hera_sim/visibilities/conversions.py
+++ b/hera_sim/visibilities/conversions.py
@@ -33,7 +33,7 @@ def uvbeam_to_lm(uvbeam, freqs, n_pix_lm=63, **kwargs):
     az = -np.arctan2(m, L)
     za = np.pi / 2 - np.arcsin(n)
 
-    efield_beam = uvbeam.interp(az, za, freqs, **kwargs)[0]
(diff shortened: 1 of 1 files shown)


steven-murray (approved):
Thanks @aewallwi

PR #162, opened 2021-07-15
Initial changes for simulator overhaul

steven-murray:
This branch changes the structure of the VisibilitySimulator to have three classes: 

1. The `ModelData` class to hold the configuration for a simulation (UVData, UVBeam and SkyModel)
2. The `VisibilitySimulator` class, which should be subclassed to define particular simulators.
3. The `VisibilitySimulation` class which ties them together.

Together, these provide a much more flexible and clean interface for defining new simulators, and using them.

We've also updated the constructor for the `ModelData` (which used to be the `VisibilitySimulator`) so that it can be cleanly constructed from a pyuvsim config file.


r-pascua:
Err, do you know why the entire commit history from mid-v1.0.0-dev to now is part of this PR's commit history? It's saying that almost every, if not every, file has been changed in this PR. Do you know if this will be a non-issue when this is merged into `main`?

steven-murray:
darn. Sorry, see #165

PR #163, opened 2021-07-15
update-sim-classes

steven-murray:
This depends on the initial PR. 

It fixes `VisCPU` to use the new setup.

@hughbg can you check especially if the call to `correct_point_source_pos` is good? And also if polarization is being used correctly?


steven-murray:
@hughbg  and @r-pascua I think this should be ready for a good review now -- the tests are all working, at least locally.

r-pascua:
I just want to quickly raise the same concern I did on the other PR. Was the `visibility-sim-overhaul` perhaps branched from a pre-v1 branch? I have two main concerns when bringing this up: first, I'm worried that the commit history will get very messy (which might actually be a non-issue); second, it's a lot easier to do a comprehensive review when only the changes that were actually made are listed as the changes.

steven-murray:
@r-pascua yeah, I branched out from master instead of main 🤦🏻 . I'll try closing and re-opening the PR...

PR #164, opened 2021-07-19
kwargs and memo ref added

samirchoudhuri:


hera_sim/tests/test_polybeam.py
```diff
@@ -179,3 +179,8 @@ def test_perturbed_polybeam(self):
         # Check that attempting to use GPU with MPI raises an error.
         with pytest.raises(RuntimeError):
             run_sim(r, use_gpu=True, use_mpi=True)
+
+
+def test_polybeam():
```
steven-murray, line 184: try to make the test name a bit more informative: maybe `test_polybeam_kwargs`

hera_sim/beams.py
```diff
@@ -25,7 +25,7 @@ class PolyBeam(AnalyticBeam):
         Reference frequency for the beam width scaling power law, in Hz.
     """
 
-    def __init__(self, beam_coeffs, spectral_index=0.0, ref_freq=1e8):
+    def __init__(self, beam_coeffs, spectral_index=0.0, ref_freq=1e8, **kwargs):
```
steven-murray, line 28: If you add `**kwargs`, you'll need to document where exactly those arguments get sent. As a user, I can see that it takes `**kwargs`, but I don't know what arguments I can actually pass or what they do.

steven-murray (commented):
Thanks @samirchoudhuri. Just a couple of comments to address.

hera_sim/tests/test_polybeam.py
samirchoudhuri: Thanks @steven-murray . I have modified the function name.

hera_sim/beams.py
samirchoudhuri: Here, we used  **kwargs in PolyBeam to use the same dictionary input format as for PerturbedPolyBeam class.

steven-murray: But you don't need it. Those `**kwargs`  aren't being used for anything. I can't see a use-case in which you require them to be in the base class? THe kwargs in `PerturbedPolyBeam` get passed through to `PolyBeam`, which makes sense. But the ones in `PolyBeam` don't do anything. You might as well not have them.

steven-murray (approved):
THanks @samirchoudhuri !

PR #167, opened 2021-07-20
Add polarized PolyBeam to viscpu_pol

WarmongeringBeaver:
Includes code for the polarized version of `PolyBeam`, and associated unit tests.


hera_sim/beams.py
```diff
@@ -121,41 +323,48 @@ def interp(self, az_array, za_array, freq_array, reuse_spline=None):
...
+
         interp_basis_vector : array_like
             Array of interpolated basis vectors (or self.basis_vector_array
             if az/za_arrays are not passed), shape: (Naxes_vec, Ncomponents_vec,
             Npixels/(Naxis1, Naxis2) or az_array.size if az/za_arrays are passed)
         """
+        # Check that coordinates have same length
+        if az_array.size != za_array.size:
+            raise ValueError(
+                "Azimuth and zenith angle coordinate arrays must have same lentgh.")
```
steven-murray, line 335: Typo in length

steven-murray (commented):
Thanks @WarmongeringBeaver. I'm generally happy with it but have a couple of questions:

1. For the functions adapted from pyuvdata, could we not have just used the functions in pyuvdata? Or was there a lot of 'adaption'?
2. Should this be merged into viscpu_pol or main? There is no open PR for viscpu_pol, and it looks like viscpu_pol is out of date with the main code.

WarmongeringBeaver:
1. There were mainly checks and options removed, but the core code is identical. Directly using the methods of `pyuvdata.UVBeam` would however require to convert an evaluated `AnalyticBeam.PolyBeam` into a `UVBeam`, or more generally an evaluated `pyuvsim.AnalyticBeam` into a `UVBeam`. If there are other cases where this conversion would be useful, then why not! 
2. `viscpu_pol` is indeed behind `main`; I don't know if it's less cumbersome to merge to `viscpu_pol` first and then update it (and maybe delete it), or merge to `main` directly.

steven-murray:
It seems like the core code should be refactored in pyuvdata then, so it can be used directly, rather than doing some instantation. But that's for another time...

steven-murray:
I think I'm worried that viscpu_pol is no longer required in hera_sim since `vis_cpu` became a thing. I can't quite remember what `viscpu_pol` did. If it's still definitely required, then yes, lets' merge this into `viscpu_pol`, then merge `main` into there.

steven-murray:
I looked at `viscpu_pol` and it is definitely out of date. You might want to work with @philbull to ensure the right things are being merged.

philbull:
I am working through the merge conflicts now

WarmongeringBeaver:
The tests seem to be failing because `healpy` is imported (by the test scripts only, not in `beams.py`) but not installed.

steven-murray:
@WarmongeringBeaver you should try to use `astropy_healpix` instead of healpy (for licensing reasons). You can add it to the `dev` and `tests` extra deps lists in `setup.py`

WarmongeringBeaver:
Needs #https://github.com/HERA-Team/vis_cpu/pull/20 to pass now.

philbull:
Hi folks, note that the `viscpu_pol` branch has a number of recent changes, including a fix to some PolyBeam normalisation bugs that will affect this PR. I also made some minor test improvements and fixed some floating point handling.

steven-murray (approved):
Thanks @WarmongeringBeaver, this is fine work. 

I'll leave it up to you whether you want to add more tests now, or before merging the `viscpu_pol` branch into main. Otherwise, feel free to click the shiny green button!

PR #168, opened 2021-07-20
fix: sky model from obsparams was transposed for point sources

steven-murray:
Fixes the transposition bug for point source input.


r-pascua (approved):
Cool, I think this should do it. Thanks!

PR #169, opened 2021-07-20
fix: simple fix to beam ids in vis_cpu

### LINKED ISSUE #137: Choosing antenna beams is kinda buggy
beam handling is slightly buggy; it's possible to input `beams=None` and `beam_ids=[0, 1, 2]`, which will cause an error later on. May not be checked fully.

steven-murray:
Fixes #137 

This *should* fix the beam ids problem, but give it a quick test first.

### DIFF
--- a/hera_sim/tests/test_vis.py
+++ b/hera_sim/tests/test_vis.py
@@ -738,5 +738,30 @@ class TestSimRedData(unittest.TestCase):
                 np.testing.assert_almost_equal(ans0yy, ans_yy, decimal=7)
 
 
+def test_nants_beams(uvdata2):
+    beams = [AnalyticBeam("uniform"), AnalyticBeam("gaussian")]
+
+    freqs = np.unique(uvdata2.freq_array)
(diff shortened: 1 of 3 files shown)


philbull:
@steven-murray This seems to change an intentional behaviour, which was to interpret the passing-in of a single beam as saying "use this beam for all antennas". If that behaviour is no longer desired, the tests need to be updated.

steven-murray:
@philbull on the contrary, that is the default behaviour of the `VisibilitySimulator` class -- `VisCPU` was trying to re-implement the same behaviour it was already inheriting. This PR just lets the super class do the work.

steven-murray:
@r-pascua I think this is now ready. The patch coverage is failing, and I'm not sure why, but it's literally from a single line not being covered.

r-pascua (approved):
OK, I tried running the same things that (I think) this was failing on before, and it seems to work. This also looks like the right thing to do. Thanks for fixing this!

PR #171, opened 2021-07-27
Zernike beam model

philbull:
Add an analytic beam model based on 2D Zernike polynomials.


hera_sim/beams.py
```diff
@@ -249,3 +249,741 @@ def interp(self, az_array, za_array, freq_array, reuse_spline=None):
...
+
+        return interp_data, interp_basis_vector
+
+    def __eq__(self, other):
+        """Evaluate equality with another object."""
+        if not isinstance(other, self.__class__):
+            return False
+        return self.beam_coeffs == other.beam_coeffs
+
+    def zernike(self, coeffs, x, y):
```
steven-murray, line 349: Looks like this could be a `@staticmethod`?

hera_sim/beams.py
```diff
@@ -249,3 +249,741 @@ def interp(self, az_array, za_array, freq_array, reuse_spline=None):
             interp_data += (1.0 - step) * (mainlobe_pert - mainlobe0)
 
         return interp_data, interp_basis_vector
+
+
+class ZernikeBeam(AnalyticBeam):
```
steven-murray, line 254: Any reason this doesn't subclass the `PolyBeam`?

hera_sim/beams.py
```diff
@@ -249,3 +249,741 @@ def interp(self, az_array, za_array, freq_array, reuse_spline=None):
...
+            Values of the Zernike polynomial at the input x,y points.
+        """
+        # Coefficients
+        assert len(coeffs) <= 66, "Max. number of coeffs is 66."
+        c = np.zeros(66)
+        c[: len(coeffs)] += coeffs
+
+        # x2 = self.powl(x, 2)
+        # y2 = self.powl(y, 2)
+        x2, x3, x4, x5, x6, x7, x8, x9, x10 = (
```
steven-murray, line 376: maybe rewrite as `xpow = tuple(x**idx for idx in range(2, 11))`

hera_sim/beams.py
```diff
@@ -249,3 +249,741 @@ def interp(self, az_array, za_array, freq_array, reuse_spline=None):
...
+            x ** 3.0,
+            x ** 4.0,
+            x ** 5.0,
+            x ** 6.0,
+            x ** 7.0,
+            x ** 8.0,
+            x ** 9.0,
+            x ** 10.0,
+        )
+        y2, y3, y4, y5, y6, y7, y8, y9, y10 = (
```
steven-murray, line 387: same as above

hera_sim/beams.py
```diff
@@ -249,3 +249,741 @@ def interp(self, az_array, za_array, freq_array, reuse_spline=None):
...
+            y ** 6.0,
+            y ** 7.0,
+            y ** 8.0,
+            y ** 9.0,
+            y ** 10.0,
+        )
+
+        # Setting the equations for the Zernike polynomials
+        # r = np.sqrt(powl(x,2) + powl(y,2))
+        Z1 = c[0] * 1  # m = 0    n = 0
```
steven-murray, line 401: formatting-wise, I think it might be easier to set z as a dict, like

```python
z = {
    1: c[0] * 1,
    2: ...
}
```

Then the final thing can just be `sum(z.values())`. The above formatting makes it easy to identify a particular coefficient still.

hera_sim/tests/test_zernikebeam.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+def beams(nants):
+    """
+    Zernike polynomial beams with some randomly-chosen coefficients.
+    """
+    cfg_beam = dict(
+        ref_freq=1.0e8,
+        spectral_index=-0.6975,
+        beam_coeffs=[
```
steven-murray, line 47: these one-per-line expansions by black on lists of floats are kind of annoying. You can avoid it by using `# nofmt` on the first line of the list.

hera_sim/tests/test_zernikebeam.py
```diff
@@ -0,0 +1,148 @@
...
+        precision=2,
+    )
+    simulator.simulate()
+
+    auto = np.abs(simulator.uvdata.get_data(0, 0, "XX")[0][0])
+
+    return auto
+
+
+class TestZernikeBeam:
```
steven-murray, line 136: don't require a class here.

steven-murray (commented):
Looking good so far, @philbull. Give me a ping when it's ready for a final review!

hera_sim/beams.py
philbull: Yes, it implements a 2D polynomial basis instead of 1D + various transformations (rotation/stretch/etc), so there's not really anything to be gained from subclassing.

hera_sim/beams.py
philbull: Much more elegant, thanks

steven-murray:
Ah, as usual I jumped the gun with the approval -- please update the README!

steven-murray:
@philbull tests still seem to be failing -- it might be that it didn't pick up the latest version of vis_cpu.

PR #172, opened 2021-07-29
Enable polarisation handling for vis_cpu

philbull:
Enables polarisation handling capabilities for vis_cpu. This currently allows for polarised beams, but only an unpolarised (Stokes I) sky model.

A polarised version of the analytic `PolyBeam` beam model is also included.


steven-murray:
@philbull looks like there's some conflicts with `main`. I'll be happy to review once those are sorted!

philbull:
I've added a test that compares the vis_cpu and pyuvsim outputs for polarised and unpolarised beams. Because of the `mpi4py` dependency of `pyuvsim`, I've had to change the unit test system to use miniconda.

These comparison tests should start passing once https://github.com/HERA-Team/vis_cpu/pull/21 has been merged.

hera_sim/beams.py
```diff
@@ -11,26 +259,46 @@ class PolyBeam(AnalyticBeam):
...
     angles when the beam is interpolated, using a power law.
 
-    See HERA memo
+    See HERA memo:
     http://reionization.org/wp-content/uploads/2013/03/HERA081_HERA_Primary_Beam_Chebyshev_Apr2020.pdf
 
     Parameters
     ----------
-    beam_coeffs : array_like
+    beam_coeffs: array_like
```
steven-murray, line 267: there does need to be a space before the colon here

hera_sim/beams.py
```diff
@@ -104,21 +388,27 @@ def __eq__(self, other):
 
 
 class PerturbedPolyBeam(PolyBeam):
-    """
-    A PolyBeam in which the shape of the beam has been modified.
+    """A PolyBeam in which the shape of the beam has been modified.
```
steven-murray, line 391: for documentation cross-referencing, it would be great if you could use 
```
:class:`PolyBeam`
``` 
here instead of just the name.

hera_sim/beams.py
```diff
@@ -104,21 +388,27 @@ def __eq__(self, other):
...
+
+    Note that the PolyBeam `beam_coeffs` kwarg must be passed on instantiation.
```
steven-murray, line 393: Can we make this explicit in the signature?

hera_sim/beams.py
```diff
@@ -208,7 +611,7 @@ def interp(self, az_array, za_array, freq_array, reuse_spline=None):
 
             # Updated polar coordinates
             theta_s = np.sqrt(Xs ** 2.0 + Ys ** 2.0)
-            phi_s = np.arccos(Xs / theta_s)
+            phi_s = np.where(theta_s == 0.0, 0.0, np.arccos(Xs / theta_s))
```
steven-murray, line 614: This is still going to throw up annoying warnings of divide by zero. Can you make it two lines and actually mask the calculation?

hera_sim/tests/test_compare_pyuvsim.py
```diff
@@ -0,0 +1,331 @@
...
+
+
+def test_compare_1src_polybeam_pol():
+    # Single source, PolyBeam beam, polarized
+    compare_viscpu_with_pyuvsim(nsource=1, beam_type="PolyBeam polarized")
+
+
+def test_compare_100src_gauss():
+    # Single source, Gaussian beam, unpolarized
+    compare_viscpu_with_pyuvsim(nsource=100, beam_type="gaussian")
```
steven-murray, line 321: It is more idiomatic to write these as parameterizations, so instead of:

```python
def compare_viscpu_with_pyuvsim(nsource, beam_type):
```

you'd have

```python
@pytest.mark.parameterize('nsource,beam_type', [(1, 'gaussian'), (1, 'PolyBeam'), ...])
def test_viscpu_vs_pyuvsim(nsource, beam_type):
```

hera_sim/tests/test_polybeam.py
```diff
@@ -179,3 +195,178 @@ def test_perturbed_polybeam(self):
...
+        assert np.all(~np.isinf(calc_result_nn))
+        assert np.all(~np.isnan(calc_result_en))
+        assert np.all(~np.isinf(calc_result_en))
+        assert np.all(~np.isnan(calc_result_ne))
+        assert np.all(~np.isinf(calc_result_ne))
+        assert np.all(~np.isnan(calc_result_unpol))
+        assert np.all(~np.isinf(calc_result_unpol))
+
+
+class TestPolarizedPolyBeam:
```
steven-murray, line 224: Looks like the class is not necessary here

hera_sim/tests/test_vis.py
```diff
@@ -386,7 +386,7 @@ def test_comparison_zenith(uvdata2):
     ).simulate()
 
     assert viscpu.shape == healvis.shape
-    np.testing.assert_allclose(viscpu, healvis, rtol=0.05)
+    np.testing.assert_allclose(viscpu.conj(), healvis, rtol=0.05)
```
steven-murray, line 389: We shouldn't have to take the conj here, I don't think. I'm pretty sure work on #165 will remedy this.

steven-murray (commented):
This is really great, @philbull and @WarmongeringBeaver. I've made a few small suggestions, but the main thing I want to put out there is that #165 is in progress, and it tries to streamline the API of the `VisibilitySimulator` class. There is also a PR out (#170 ) for adding a direct wrapper for pyuvsim.

The reason I mention these is firstly because several parts of this code will have to be modified to suit that API very soon, but also I *think* some parts of this code will be much neater after those PRs are merged. For instance, in the comparison test, much of the creation of the `VisCPU` class will be done automatically in the new API. There are also more careful checks of baseline ordering and conjugation convention in that PR, which should hopefully help. 

So, the question is -- do we want to merge this to master now, then pull it back up to #165 and deal with the conflicts there, or do we first want to merge #165, and update this PR to use those changes?

hera_sim/tests/test_vis.py
philbull: I'll leave it as-is for now

steven-murray (approved):
Thanks @philbull. My only suggestion is to get the patch coverage  up to where it should be -- probably requires one or two small new tests.

PR #173, opened 2021-08-03
Blt ordering

r-pascua:
This PR makes sure that the healvis wrapper output is sorted correctly.

### DIFF
--- a/hera_sim/visibilities/healvis_wrapper.py
+++ b/hera_sim/visibilities/healvis_wrapper.py
@@ -167,4 +167,4 @@ class HealVis(VisibilitySimulator):
 
             vis[indx] = vis_here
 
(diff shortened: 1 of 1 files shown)


steven-murray (approved):
Well, I'm not sure I'm following the logic perfectly, but if the tests pass, I think we're in good shape!

r-pascua:
@steven-murray I added some comments explaining the logic, can you let me know if they help you understand why it works? If they're clear enough, then I'll merge, and we can fix the failing test in #165 or something. (We just need to cover the case of the user providing a list for the ``beam_ids``, assuming that it's one ID per antenna.)

steven-murray (approved):
Thanks @r-pascua, yes that is much more understandable :-) I'm still not sure why it now works if the times have been reordered, but if tests pass, tests pass :-)

PR #175, opened 2021-08-26
More specific checks on polarization inputs for vis_cpu

steven-murray:
This PR attempts to make the dealing of vis_cpu with polarization a bit more robust.

It adds an input option that allows the user to turn off raising errors if not all the polarizations in the UVData object are going to be filled (this might not be a good idea, but the idea was that one might want to do `polarized=False` with a full 4-pol object, and just have it add to the XX, say). 

It also sets a dynamic default for `polarized` that looks if the input UVData object has multiple polarizations or not. Again, not exactly sure this is the best idea -- perhaps it should explicitly check if the only pol in the object is 'XX'.

It also sets the available polarizations (in the output of vis_cpu) when NOT doing polarization to _either_ `ee` or `nn` depending on the `x_orientation`. Not sure if this shouldn't just be done directly as `XX` instead of assuming anything about the x orientation at all. 

It also adds a bunch of tests to simulate with polarization on/off with different inputs, to make sure we get all the cases (with some expected to raise an error). 

Now, there's a few more things I'm not sure of:

1. When doing non-polarized, should we expect the output of vis_cpu to be 'XX', or 'pI' (and if 'pI', should it be 'XX + YY' or '(XX+YY)/2'? 
2. If the input object only has 'YY' for some reason, should we consider that to be OK and just put non-polarized sim data into that pol?
3. If the output of vis_cpu (in unpolarized case) is really definitely XX (not pI), could we convert the output if the input obj is defined in pI?

### DIFF
--- a/hera_sim/beams.py
+++ b/hera_sim/beams.py
@@ -36,7 +36,7 @@ def stokes_matrix(pol_index):
     elif pol_index == 3:
         pauli_mat = np.array([[0.0, -1.0j], [1.0j, 0.0]])
     else:
-        raise ValueError("'pol_index' most be an integer between 0 and 3")
+        raise ValueError("'pol_index' must be an integer between 0 and 3")
 
     return pauli_mat
 


hughbg:
Some more thoughts: What would probably make sense to the user is that they specify in uvdata what pols they want, and that’s what they get. The polarized keyword wouldn’t be necessary because they've specified everything in uvdata.  

I think that XX etc.should mean the same as they do for pyuvsim, so pI is XX+YY.

If “pI” was specified as the pol for hera_sim, does it work? Is pI fully implemented?

hughbg:
I ran some tests using a batch script that I've attached (couple of scripts plus the output in the log). It tries lots of different configurations. If an error is generated because some pols won’t be filled, the script retries the simulation with allow_empty_pols=True. Everything looks good. 

[tests.zip](https://github.com/HERA-Team/hera_sim/files/7079114/tests.zip)

piyanatk:
My general thought is whatever is the behaviour in `pyuvsim`, we follow.

@steven-murray on your specific questions
> 1. When doing non-polarized, should we expect the output of vis_cpu to be 'XX', or 'pI' (and if 'pI', should it be 'XX + YY' or '(XX+YY)/2'?
I think it should be `['XX', 'YY']` in the `polarization_array`. My though is the visibility produced by a simulator should produce what come out from the correlator, and that is never 'pI'

> 2. If the input object only has 'YY' for some reason, should we consider that to be OK and just put non-polarized sim data into that pol?
I think we raise error - it should be the user job to provide both 'XX' and 'YY'. Though I am a little confused here. Sky model, e.g. as a `pyradiosky` object, are in Stokes right? So what else actually need to be in 'XX' and 'YY' apart from beam?

> 3. If the output of vis_cpu (in unpolarized case) is really definitely XX (not pI), could we convert the output if the input obj is defined in pI?
I am not sure I understand this question. But I think both the input and output UVData object should be defined in ['XX', 'YY', 'XY', 'YX']. All other conversation to pseudo-Stokes, except those needed internally, should be users' tasks.

steven-murray:
Thanks @piyanatk. 

What I'm getting from you is that we should force the input UVData to have `polarization_array` as either `['XX']` or `['XX', 'YY', 'XY', 'YX']`. In the former case, we don't do polarization in vis_cpu, in the latter we do. Any other input should be flagged as something we don't support. Is that right? 

Can I call on @mkolopanis or @bhazelton to second this? Note that currently vis_cpu doesn't deal at all with polarized sky models, just polarized beams (though it will in the future). I'm guessing that we should get the relevant polarizations from the UVBeam input object. What happens if the input UVBeam object is a power beam?

bhazelton:
I have to say I'm a bit confused about all this, I did have a look through this PR yesterday but I'm not sure I follow well enough to be able to add much. I did note some reference about the "default" value of the `x_orientation` parameter and I'd say that there is no such default, the default is to not have it set. This is because it's different per telescope (and in HERA it differs per telescope stage) so there is no sensible default.

As for the polarization array, I think this depends on vis_cpu's behavior. Power beams can generally be used with unpolarized skies to generate xx or yy visibilities but not xy or yx visibilities (I think). If you have polarized sources I think you need the full Efield beams.

bhazelton:
I do agree that it seems cleanest to have simulators create instrumental polarization visibilities not pseudo-stokes visibilities.

steven-murray:
Relevant input stuff:

- UVData object:
  - polarization_array
  - x_orientation
- UVBeam object:
  - Nfeeds
  - beam_type ('power', 'efield')
- keywords user:
  - polarized=True/False

james: try to simulate whatever's in the UVData object. Only problem is if UVBeam object doesn't have necessary info. Eg. UVBeam has only powerbeam. 

What does 'power beam' really mean anyway? Does it mean pI or XX? Ask Bryna/look at docs. 

Remove keyword.

Cases:

1. Beam_type=efield, pol_array=any subset of ['xx', 'xy', 'yx', yy'] : can do it, by downselecting
2. Beam_type=efield, pol_array=psudo-stokes: can do it, but raise a (useful) error because the user should do it themselves
3. Beam_type=efield but only one feed: raise error because we can't do it (yet)

Bobby:
1. Need to be able to construct ALL the pols in the UVData object from those in the beam. ie. anything between 1-4 pols in UVData and an efield beam.
2. Have a power beam with two feeds (check if this is possible): then we can do either XX or YY or both (or is it psuedo-stokes?)
3. Have a power beam with one feed: can do either XX or YY (whichever is actually in the beam)

@bryna: what's the intended use-case for power beam? Should we just disallow it? Could allow it for an UNPOLARIZED sky. Definitely don't allow power beam with four components.

x_orientation:

1. if both not given, set both x to east, raise warning
2. if either given but not the other, set the other to the same (but raise a warning)
3. if both given and different, raise error. 
4. vis_cpu understands X as east (so have to do something if it's N)

piyanatk:
Another though on `x_orientation`, since this is `"hera"_sim`, maybe we should enforce `x` to east? But the proposed checks are good too and allow more flexibility ...

steven-murray:
To formalize the previous comment (which was made in fits and starts on a call), here is some suggested psuedo code for checks on polarizations for everyone's perusal. Note that this is specifically for vis_cpu, which (I'm pretty sure, @philbull correct me if I'm wrong) assumes that 'X' is East in terms of the input sky model. Another thing of note is that vis_cpu has two modes polarized=True or False. When polarized it always computes all four linear pols (these can be subselected after they are returned, if necessary). If unpolarized, it computes essentially an XX pol for a single feed.

OK here goes:

```python
def call_vis_cpu(uvdata, uvbeam, sky_model):
    # Check x-orientations
    if uvbeam.x_orientation is None and uvdata.x_orientation is None:
       set_x_orientation_to_east_after_warning(uvdata, uvbeam)
    elif uvbeam.x_orientation is None:
        uvbeam.x_orientation = uvdata.x_orientation
        warn()
    elif uvdata.x_orientation is None:
        uvdata.x_orientation = uvbeam.x_orientation
        warn()
    else:
        assert uvdata.x_orientation == uvbeam.x_orientation

    # Now check that we only have linear polarizations (don't allow pseudo-stokes)
    if any(pol.startswith('p') for pol in uvdata.polarization_array):
        raise ValueError("Nice error telling user to convert their uvdata object to linear pols first")

    if any(pol not in ['xx', 'xy', 'yx', 'yy'] for pol in uvdata.polarization_array):
        raise ValueError("vis_cpu can only simulate linear polarizations")

    # If we have a power-beam, then we can only simulate unpolarized sky, and only XX and/or YY
    if uvbeam.beam_type == 'power':
        assert all(pol in ['xx', 'yy'] for pol in uvbeam.polarization_array)
        assert sky_model.unpolarized
        assert len(uvbeam.polarization_array) >= len(uvdata.polarization_array)  # uvdata can only have xx and/or yy

    elif uvbeam.beam_type == 'efield':
        assert uvbeam.Nfeeds==2  # can't deal with one feed, that would be wierd?

    # Now, make sure the beam gets sorted properly if we have X == North.
    if uvbeam.x_orientation == 'n':
        uvbeam = transpose(conjugate(uvbeam))  # yeah, not sure of the exact code here

    # Decide whether we do the polarized or unpolarized version of vis_cpu
    # Should be unpolarized if we only have xx or yy in the uvdata array, OR if we have power beams.
    # In the latter case, it's possible to have BOTH xx and yy -- we could then either call vis_cpu twice independently, or just set
    # polarized=True and subselect. Not sure if these really mean the same thing though?
    polarized = uvdata.polarization_array == (['xx'] or ['yy']) or uvbeam.beam_type='power'

    # Run the actual call
    vis = vis_cpu(
        polarized=polarized,
        sky_model=sky_model,
        beams = construct_lm_beams(uvbeam)
    )

    # Decide on how to put the data back into the uvdata array.
    if not polarized:
        # leave it alone, it's all good
    else:
        for pol in uvdata.polarization_array:
            get_index_in_vis_cpu_output_given_xorient(pol)
``` 

I think the outstanding questions are:

1. @bhazelton : does the above usage of the power beam make sense? Is a power-beam in XX the same as a polarized XX component?
2. If we have a power-beam with two components, should we loop over the call to vis_cpu while setting polarized=False, or (pending whether a power-beam is the same as an XX beam) just do the polarized beam and subselect?
3. @philbull: Does vis_cpu internally enforce the assumption that X = East?
4. @philbull: In the unpolarized case, is the assumption that the input beam to vis_cpu is XX?

bhazelton:
Power beams are effectively the response of the baseline, so they are the product of the E-field beam from the first antenna with the E-field beam from the second antenna. These are useful for many things besides as an input to simulators (in fact I'd argue they aren't generally what you want for simulators, there you generally want the E-field beam) and as a result they are not limited to be only xx or yy -- you can certainly make the xy and yx power beam and that can be very useful in some situations (just maybe not for simulators). So I don't think you can assert that the xy & yx pols don't appear. Or I guess you can but you should give the user a useful error message because it's not a mistake for those to be defined on a UVBeam object.

I think you can use xx and yy power beams to generate sensible xx and yy visibilities if you do not have a polarized sky. I don't think you can use an xy and yx power beam to make sensible xy and yx visibilities of unpolarized sources, but I'm not 100% sure. I know that you can't make sensible visibilities with power beams for any polarization if you have polarized sources.

I'm not sure if that answered your question. I guess you're asking if an xx power beam is the same as the xx response to the sky and I think it is if you do not have polarized sources.

As far as UVBeam is concerned, you are allowed to only have one feed, but then you only have one polarization of visibilities of course.

I do not think you can transpose and conjugate to fix the x_orientation, that is definitely incorrect. Instead you should swap the data between the two feeds for an E-field beam, a power beam is more complicated but we can work that out if needed.

If vis_cpu can do a single power beam simulation (I assume that's what is meant by "unpolarized"?) you could imagine running it twice, once with the xx beam and the second time with the yy beam.

I'm not sure if that addressed all the relevant points, let me know if you need more.

r-pascua:
@steven-murray, the pseudocode you suggest generally seems like the right thing to do (we'll deal with modularization and getting everything right when it comes time for that). When this is cast into actual code, we're going to need to be careful about how the beams are set up and how the output from vis_cpu is ordered (e.g. ensuring that the polarization matches between the beams and the visibilities, as well as getting the filtering step right in the odd case of polarized data with less than four pols). I think we'll also want to add a check enforcing that the beam hasn't been converted to pStokes before running the simulation.

@bhazelton's response brought a few things to mind. First, I think that we should check that an efield beam is used when the sky model is polarized. On a somewhat related train of thought, I'm thinking that instead of comparing the lengths of the beam and data polarization arrays, we should check that the data polarization array is a subset of the beam polarization array. Finally, I'm thinking that we may be better off disallowing power beams entirely for the vis_cpu wrapper&mdash;if I'm understanding how vis_cpu works correctly, then it actually expects to get an efield beam (since the calculation is vectorized over antennas), so we'd have to modify the beam before simulating visibilities, but I think that this could lead to unexpected problems at a post-simulation analysis stage (and so we should save ourselves/others the headache). Please do let me know if I'm speaking nonsense, though, as I'm sluggishly ramping back up to being in science-mode after a few weeks of moving.

I might have some thoughts regarding item 2 in the list, but I'll hold off on that until after further discussion regarding whether we allow power beams at all in light of what they are meant to represent.

On a different, but related, note: I still think we should add an attribute that tracks the pI convention (that is, whether pI = XX + YY or pI = (XX + YY) / 2), and add a bit of logic that makes the simulated visibilities agree with the convention used by pyuvsim.

piyanatk:
I second @r-pascua's comment on disallowing power beam entirely if the computation in VisCPU expect e-field beam.

On a related note, are we currently checking the polarization by using the string labelling, e.g. `['xx', 'xy', 'yx', 'yy']`? I believe UVData exclusively use the AIPS integer labelling, that is, `['xx', 'xy', 'yx', 'yy'] = [-5, -6, -7, -8]` and `['pI', 'pQ', 'pU', 'pV'] = [1, 2, 3, 4]`. Perhaps, we should do all internal polarization check with the integer labels.

steven-murray:
@bhazelton thank you for your helpful response!

OK, I think I'm getting it now. I think what we have been doing for power beams so far is taking the sqrt to apply to each antenna. I guess this makes sense in the limited case that all antennas have the same beam and the sky is unpolarized. Though this is a limited case, it seems to me to be a case that might be used a lot for more exploratory simulations. For example, using an AnalyticBeam of some variety I expect to be defined as a power beam. So I reckon that we should support this case.

We'll have to work out the details for when the `x_orientation` is North. @bhazelton when you say "swap the data between the feeds" which data do you mean? For the beam or the visibilities? 

@r-pascua: good idea, I'll add a check to ensure the beam isn't in pstokes. I think I already had a check that the beam is efield if the sky is polarized. I considered making the data pol array a subset of the beam one, but because you can swap the X/Y potentially depending on the x-orientation, this is a bit more complicated. 

On your last point (about pI): I think the plan is that vis_cpu will never touch pstokes, right? We'll always deal in linear pols. So do we still need to track that? I think pyuvsim also deals in linear pols.

bhazelton:
For the x_orientation, you should swap the data in the arrays for either or both the UVData and UVBeam objects depending on their values compared to what it assumed by the simulator. If the simulator makes no assumption (and I don't think it needs to) about x_orientation, then the two objects need to agree and if they don't you should swap the data for one of them to make them agree. If you're working with E-field beams, it's probably easier to swap the data for the UVBeam object. For a power beam or a UVData object it's more complicated because you have products of the feeds so you need to work out how to swap them (I haven't thought about it deeply -- it might be simply data swapping but it might also require some data conjugation).

Effectively, the `x_orientation` parameter is just a label -- it says whether the `x` feed is aligned N/S or E/W. So if you want to make the objects agree you can just swap the data between the feeds and change the value of the `x_orientation` label. We could even make a method on the UVBeam and/or UVData objects to do that swapping to make sure it's done correctly.

hughbg:
@philbull: Does vis_cpu internally enforce the assumption that X = East?

I don't see anything in the code to do with orientation as defined by x_orientation in vis_cpu. It expects everything to be setup right beforehand. However there is a complication. When we were putting the astropy code into vis_cpu we discovered that pyuvsim does a correction to apply the UVBeam az/za convention. This rotates source azimuths, and is [done here](https://github.com/RadioAstronomySoftwareGroup/pyuvsim/blob/f6bdf5581b78c5f412463ca21d2d3b02d85556e7/pyuvsim/utils.py#L70). I don't know how this interacts with the x_orientation. The ability to apply this convention to vis_cpu was added in the function [enu_to_az_za()](https://github.com/HERA-Team/vis_cpu/blob/687cb7c92e30b5aee4b094304e63753c8966d63f/src/vis_cpu/conversions.py#L10), which calculates az/za.  In the docstring for that function you can see there are certain assumptions about orientation..

@philbull: In the unpolarized case, is the assumption that the input beam to vis_cpu is XX?

In the original hera_sim, which this case conforms to, the beam is unpolarized and has real values. The visibilities are simulated just based on source locations through that beam and then the results are put in the "XX" slot in the output. The source locations are given to vis_cpu via eq2tops, crd_eq, but then converted to az/za  using enu_to_az_za().

bhazelton:
There is no interaction between the coordinates of the beam and sky, which is what the conversion between the az/el and az/za coordinates are about, and the `x_orientation`. The `x_orientation` is just a label for which feed gets put into the data arrays in the `x` slot as opposed to the `y` slot.

steven-murray:
So @bhazelton, just to be clear -- a particular beam knows how to interpolate itself onto az/el in an absolute sense regardless of its x_orientation? As in, changing it's x_orientation to "north" from "east" will rotate its beam appropriately when calling `.interp()`? Or am I being daft?

bhazelton:
I think you are confused about what `x_orientation` means. It's just a label. It has no interaction with the `interp` method.

steven-murray:
OK, I think this is now ready for review, @r-pascua and @bhazelton  (and @philbull if you are off vacation!). 

The final idea was to be more conservative in what we support, at least initially. This means we don't support power beams (because vis_cpu does not yet support them) and we also *force* the x_orientation to be 'north' (i.e. we error if it is not north), which is something that is *assumed* in vis_cpu, at least for the unpolarized case (that is, when getting beam in lm co-ordinates, it takes the (phi, y) part of the efield beam, which it comments to be "east east"). With these limitations, the code is much easier to write. 

It would be good in the future to support more cases, but we'll have to be careful about it, and assess the demand for it.

bhazelton:
Can you elaborate on this:
> (that is, when getting beam in lm co-ordinates, it takes the (phi, y) part of the efield beam, which it comments to be "east east")
I'm worried that there's still confusion about what the `x_orientation` parameter means. Maybe it would be useful to talk this through on zoom sometime?

steven-murray:
@bhazelton yes, I am definitely still confused about the x_orientation parameter. I think the code now is "correct" but probably not as general as it could be. I'd be very happy to zoom about it!

steven-murray:
@r-pascua and @philbull I think this is now finalized, but it will depend on https://github.com/HERA-Team/vis_cpu/pull/29 being merged first. 

@r-pascua would appreciate a review, especially of how the indexing into the final output is done.

hera_sim/visibilities/vis_cpu.py
```diff
@@ -356,50 +353,68 @@ def simulate(self, data_model):
...
+            y = uvbeam.feed_array.tolist().index("y")
+
+            avail_pols = {"xx": (x, x), "xy": (x, y), "yx": (y, x), "yy": (y, y)}
+            req_pols = []
+            for pol in uvdata.polarization_array:
+                # Get polarization strings in terms of n/e feeds
+                polstr = pyuvdata.utils.polnum2str(pol).lower()
+                req_pols.append(avail_pols[polstr])
+
+            return req_pols
```
r-pascua, line 408: Based on the `UVBeam` docs, it looks like this isn't compatible with the full features of the class. This is fine if we're going to require users to have their beam objects only use `'x'` and `'y'` as feed labels, but maybe we want to be loud about it (e.g. issue a warning or raise an error) if that's what we're going to do. I've got an inkling of how to do this in a generalized way, so let me know if you want me to give it a try.

r-pascua (approved):
This is looking great! Thanks for all your hard work! I have just a few comments, but nothing super big. I suspect that this should all work out once the `allow-power-beams` branch is merged in `vis_cpu`&mdash;is that right?

hera_sim/visibilities/vis_cpu.py
```diff
@@ -241,28 +280,21 @@ def get_eq2tops(self, uvdata: UVData, lsts: np.ndarray):
             dtype=self._real_dtype,
         )
 
+    def get_feed(self, uvdata) -> str:
+        """Get the feed to use from the beam, given the UVData object."""
```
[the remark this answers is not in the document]
steven-murray: Good point, I'll update the docstring.

hera_sim/visibilities/vis_cpu.py
```diff
@@ -356,50 +353,68 @@ def simulate(self, data_model):
...
-                        vis_here = np.conj(vis_here)
+        return visfull
 
-                    visfull[indx, 0, i, p] = vis_here
+    def _reorder_vis(self, req_pols, uvdata, visfull, vis, ant_list, polarized):
+        indices = (
+            np.triu_indices(vis.shape[3])
+            if polarized
+            else np.triu_indices(vis.shape[1])
+        )
```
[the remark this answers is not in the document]
steven-murray: Ah, yes.

hera_sim/visibilities/vis_cpu.py
steven-murray: Yeah, I was worried about that. We should get it right here, so if you have an idea of how to do it generally, go ahead!

r-pascua: Here's what I'm thinking:
```
if not polarized:
    return [(0,0)]
else:
    feeds = list(uvbeam.feed_array)
    # in order to get all 4 visibility polarizations for a dual feed system
    vispols = set()
    for p1, p2 in itertools.combinations_with_replacement(feeds, 2):
        vispols.add(p1 + p2)
        vispols.add(p2 + p1)
    avail_pols = {
        vispol: (feeds.index(vispol[0]), feeds.index(vispol[1])) for vispol in vispols
    }
    # Get the mapping from uvdata pols to uvbeam pols
    req_pols = list(
        avail_pols[uvutils.polnum2str(polnum, uvbeam.x_orientation)]
        for polnum in uvdata.polarization_array
    )
    return req_pols    
```
If you suspect issues with the case of the feed labels and visibility polarizations, then we could throw in some `lower` methods, but I think this should work. The output should be a list with the same order as the data polarization array giving the indices to the corresponding feeds in the beam object.

steven-murray: Hold on -- doesn't `polnum2str` only ever return something in `[xx, xy, yx, yy]`? What is the use-case you're thinking this covers?

r-pascua: Only if the `x_orientation` parameter isn't specified. It's kind of hard to tell from the `pyuvdata` source code, but testing with version 2.1.5 shows that setting an `x_orientation` will output something in `['nn', 'en', 'ne', 'ee']`. The added bonus for this is that the data object doesn't care about what the beam object says its `x_orientation` is, since we're going from polarization numbers (which are `x_orientation` agnostic) to feed labels (which could change depending on whether an `x_orientation` is specified). Put another way, we don't actually need the `x_orientation` attribute to match up between the data and beam objects in order to sort the simulated data properly (though if we're trying to do our accounting properly, then maybe we set the data `x_orientation` to match the beam's after the fact). Does that make sense?

r-pascua: I did some local testing, and this works whether the feed array uses the "xy" convention or the "en" convention. The original implementation implicitly requires that the feed array uses the xy convention. So to put it simply, this implementation covers both conventions (unless somehow the user is able to provide a beam object whose feeds are labeled with the "en" convention without specifying an `x_orientation`).

steven-murray: Ah, I see! This is great -- can you cp/paste the code in?

r-pascua: Done

PR #181, opened 2021-09-29
ci: add deprecation package to ci/tests.yml

steven-murray:
While `deprecation` is being installed in tests because we're installing the package with its deps via pip, it isn't in the `ci/tests.yml` environment. This PR adds it so that pyuvsim can install it more easily.


bhazelton (approved):
Looks good, thanks!

PR #182, opened 2021-09-30
doc: raise warning on using vispu because of baseline conjugation

steven-murray:
Adds a warning whenever using VisCPU, because it doesn't get baseline conjugation right in all cases. 

Might be better to specify exactly when it gets it wrong, but hopefully just having the warning will be enough until v2 comes.

### DIFF
--- a/hera_sim/tests/test_compare_pyuvsim.py
+++ b/hera_sim/tests/test_compare_pyuvsim.py
@@ -117,6 +117,9 @@ def test_compare_viscpu_with_pyuvsim(nsource, beam_type):
     if "polarized" in beam_type.lower():
         polarized = True
 
+    if not polarized:
+        pytest.skip("This function is broken when polarized=False until hera_sim v2")
+
     # Random antenna locations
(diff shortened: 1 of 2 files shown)


r-pascua (approved):
I think this should do, but we could make the warning a little more helpful. After messing around with things, I think it's correct to inform the user that the conjugation for `VisCPU` will only be correct if the data is conjugated to `'ant2<ant1'`and the blt order is set to `order='time', minor_order='ant2'` before anything is simulated. (There could be some nuanced thing with how the beams are sorted that I'm missing, but I think this should be correct for most cases?)

Sorry for the delay in getting this review in&mdash;I was messing around with things to figure this out, and I was majorly confused for a while (turns out the shape of a returned slice of an array depends on how many dimensions the array has?). Also, do the tests actually fail if we don't skip the unpolarized case?

steven-murray:
Thanks @r-pascua, that's incredibly helpful. I've updated the warning.

And yes, that test does fail if we don't skip it (see https://github.com/RadioAstronomySoftwareGroup/pyuvsim/pull/363)

PR #185, opened 2021-10-14
New Crosstalk Model

r-pascua:
This PR implements an improved crosstalk model, meant to better replicate crosstalk seen in H1C observations. Some of the `Simulator` code had to be modified (in particular, stuff to do with parsing the model signature and pulling parameters from the `Simulator` object as needed), but most of the changes are in the `sigchain` module.

I still need to write tests and document the new class, but I figure we could start the review process now.

Here's an example of simulated crosstalk (on top of foreground+eor visibilities) at a single time, based on the IDR2 validation sims.
example_full_array_xtalk


hera_sim/sigchain.py
```diff
@@ -488,6 +499,95 @@ def __call__(self, freqs, autovis, **kwargs):
...
+    """Crosstalk based on HERA Memo on H1C crosstalk."""
+
+    def __init__(
+        self,
+        emitter_pos=0,
+        cable_delays=None,
+        base_amp=2e-5,
+        amp_slope=-1,
+        amp_decay_base=10,
+        Ncopies=10,
```
steven-murray, line 512: can we please use `n_copies`?

hera_sim/sigchain.py
```diff
@@ -488,6 +499,95 @@ def __call__(self, freqs, autovis, **kwargs):
...
+            max_delay,
+            amp_decay_fac,
+        ) = self._extract_kwarg_values(**kwargs)
+
+        ai, aj = antpair
+        xi = antpos[ai]
+        xj = antpos[aj]
+
+        def log(x):
+            return np.log(x) / np.log(amp_decay_base)
```
steven-murray, line 560: not sure if it matters, but if  `amp_decay_base` is constant, this would be faster if it was evaluated outside the function.

hera_sim/simulate.py
```diff
@@ -888,6 +888,9 @@ def func(freqs, ants, other=None):
         model_params = {k: v for k, v in model_params.items() if v is inspect._empty}
 
         # Pull the LST and frequency arrays if they are required.
+        # TODO: update this to allow more flexibility in which arguments are
+        # looked for (this is exactly what needs to be updated to accommodate the
+        # updated crosstalk model for IDR3 validation)
```
steven-murray, line 893: pinning a TODO

steven-murray (commented):
Thanks @r-pascua, it's hard for me to grok the physics here, but the code looks good (just a couple of suggestions). Obviously still needs docs/tests. I'd also prefer if someone like @jsdillon would also review.

hera_sim/simulate.py
r-pascua: Ah, ok, the parenthetical isn't true (since I just kinda hardwired some stuff to make it work), but I still want to address the other bit. Do you think it's important enough to get this done in this PR, or should I punt this to another PR?

hera_sim/sigchain.py
r-pascua: `amp_decay_base` is indeed constant within the scope of the `__call__` method, so I can clean this up a bit. Thanks for the tip!

steven-murray:
OK cool thanks for the explanation! I wonder, for the "extra bits" that are attached to the main peak -- if they're not physically from the same source, should they be a different Model? The user can then choose to add either or both.

r-pascua:
OK, sorry for the delay, but I believe this is now ready for review. I've fleshed out the docs and written a (somewhat simple) test to make sure it works. I think I want to leave the change to the `Simulator` (in particular, how it parses class `__init__` and `__call__` methods to figure out what parameters to pass) to another PR.

@steven-murray and @jsdillon, if you could give a review when you have time, then that would be great!

hera_sim/sigchain.py
```diff
@@ -488,6 +500,188 @@ def __call__(self, freqs, autovis, **kwargs):
...
+        Base amplitude of reflection coefficient. If `amp_slope` is set to 0, then
+        this is the amplitude of all of the reflection coefficients.
+    amp_slope
+        Power-law index describing how rapidly the reflection coefficient decays
+        with distance from the receiverator.
+    amp_decay_base
+        Logarithmic base to use when generating the additional peaks in the
+        cross-coupling spectrum.
+    n_copies
+        Number of peaks to include in the cross-coupling spectrum.
```
jsdillon, line 558: Clarify that this is the number of peaks at both positive and negative delays separately.

jsdillon (approved):
Just reviewing the new `OverAirCrossCoupling` class, which is what I think you want me to look at, I just have one tiny docstring tweak. Otherwise, looks good.

If you're feeling ambitious, I'd be really great to take the code in HERA memo 104 and apply it to a full-array simulation with this code (you can do it for a single 60-integration noise-free file, I think) to see if you get back out the same emitter position, cable delays, amplitude slope, etc. that you put in. Not sure if that's strictly necessary, but I think that'd be the ultimate check of the code.

hera_sim/sigchain.py
r-pascua: Done, let me know if it's okay.

r-pascua:
@jsdillon thanks for the review! I was thinking it would be good to perform the test you suggest. I'll see about running that test when I have a bit of free time. What do you think about adding a `delay_offset` parameter?

hera_sim/sigchain.py
jsdillon: Did you push the change?

jsdillon:
> @jsdillon thanks for the review! I was thinking it would be good to perform the test you suggest. I'll see about running that test when I have a bit of free time. What do you think about adding a `delay_offset` parameter?

I explored whether there was any good evidence for a `delay_offset` in the memo and I'm not convinced it leaving it as a free parameter made the fit appreciably better. I think we can leave it out of this simulator.

hera_sim/sigchain.py
r-pascua: Oh, sorry, I guess I had a typo in the branch I pushed it to. Should be fixed now.

jsdillon: looks good to me.

r-pascua:
Ok, gotcha. Would you like me to run the simulation on an array before merging this in? I also need to do the typing for the new class parameters, so I won't merge just yet.

jsdillon:
I leave that to you. I think making that a prerequisite of pulling in this PR is a somewhat selective demand for rigor. If you feel like it would build confidence you want to have, go for it. I don't think it has to be part of the unit_tests, of course, but if you do it, definitely post plots here.

hera_sim/simulate.py
steven-murray: let's postpone to another PR.

steven-murray (approved):
Thanks @r-pascua this looks good to me!

PR #192, opened 2021-11-11
chore: move end-to-end example files, update nb

r-pascua:
This PR just moves the end-to-end example files to a subdirectory of the package's data path, as requested by @piyanatk. The relevant code in the end-to-end notebook has been updated to reflect this change.


steven-murray (approved):
this looks exactly right, thanks @r-pascua. I assume it's breaking on the BDA, so I'll merge that change back up to get the green light.

PR #194, opened 2021-11-11
test: remove bda options in cli test script

r-pascua:


steven-murray (approved):
Great, thanks @r-pascua

PR #202, opened 2022-01-14
fix: don't simulate xtalk for autos

r-pascua:
I forgot to make sure that the new cross-coupling model is only applied to cross-correlations. This has been fixed.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,6 +5,12 @@ Changelog
 v2.2.1 [2022.01.14]
 ===================
 
+Added
+-----
+- ``OverAirCrossCoupling`` now has a parameter ``amp_norm``. This lets the user
+  decide at what distance from the receiverator the gain of the emitted signal
+  is equal to the base amplitude.
+
 Fixed
 -----
 - ``OverAirCrossCoupling`` now only simulates the systematic for cross-correlations.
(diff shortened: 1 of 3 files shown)


aewallwi:
> I forgot to make sure that the new cross-coupling model is only applied to cross-correlations. This has been fixed.

Why do we have to apply this only to cross correlations? I think that the autos might be a good place to check for cross coupling effects as well.

r-pascua:
@aewallwi this model is only valid for the cross-correlations. The autos would need a different model. In particular, the cross-coupling visibility would be the sum of all of the cross-correlations, with each cross-correlation scaled by an appropriate choice of reflection coefficient. The model here would effectively rescale the input auto to (1 + eps) times itself, with eps being the reflection coefficient for the particular antenna. In the previous validation effort, we also excluded cross-coupling from the autos, since it's effectively a second-order effect.

aewallwi:
Ah! Thanks for clarifying! I thought this was for the Josaitis cross-talk model. This is definitely fine with me!

steven-murray (approved):
Perfect, thanks @r-pascua

steven-murray:
Oh add to the changelog :-)

r-pascua:
Thanks for the reminder! I will log the changes.

steven-murray (approved):

PR #203, opened 2022-01-14
fix: set is_multiplicative in ReflectionSpectrum

r-pascua:
small bugfix. will update changelog after #202 is merged


steven-murray (approved):
Woops haha. Good to go.

PR #204, opened 2022-01-16
Fix validation of beam objects.

aewallwi:
This PR fixes a validation step for vis_cpu which currently forces the user to input `efield` beams if the `xy` and `yx` pols are to be simulated. However, `vis_cpu` actually operates on power beams and should use the `xy` and `yx` power beams. As things are currently written, if the `xy` and/or `yx` visibility pols are to be simulated then the user is forced to provide an efield beam which will silently and incorrectly apply an efield beam to the sky intensity within `vis_cpu`, producing spurious results. 

I have fixed this by changing the validation step to correctly require that the cross-pol power beams are present instead. This is what should be done for an unpolarized sky.


aewallwi:
This issue was present in an older version of hera_sim. It appears to have been fixed

PR #206, opened 2022-01-17
compare earth oriented pols.

### LINKED ISSUE #205: Possible for beam x/y pols to incorrectly be compared to uvdata directional pols.
The function  `_get_req_pols` currently assumes that the feed values for a uvbeam object with non None x_orientation are set to `N` or `E`. However, there are valid `uvbeam` objects where `feed_array` contains `'x'` or `'y'` with `x_orientation` specifying the mapping from the xy labels to the directions. As written, this function will incorrectly conclude that the `nn, `ee`, `ne` and `en` pols are not present in the uvbeam object (when converted to power).

https://github.com/HERA-Team/hera_sim/blob/4c0c324ebc99297c6eef742cbf93dcec880bf372/hera_sim/visibilities/vis_cpu.py#L414

aewallwi:
Closes #205

### DIFF
--- a/hera_sim/visibilities/vis_cpu.py
+++ b/hera_sim/visibilities/vis_cpu.py
@@ -418,8 +418,19 @@ class VisCPU(VisibilitySimulator):
         # TODO: this can be updated to just access uvbeam.feed_array once the
         # AnalyticBeam API has been improved.
         feeds = list(getattr(uvbeam, "feed_array", ["x", "y"]))
-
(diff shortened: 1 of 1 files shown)


steven-murray (commented):
Thank you @aewallwi. I do believe this is correct (though it always hurts my brain to think about the polarizations/feeds between uvbeam and uvdata). 

Before approving, I'd like to see a short update to the changelog, and if possible, a test that would have failed previously.

PR #208, opened 2022-01-19
feat: option to peak-normalize beams in ModelData

r-pascua:
This PR adds a `normalize_beams` option to the `ModelData` initialization routine. If set to `True`, this ensures that all of the beams used are peak-normalized. I'm not sure if this is necessarily exactly how we want to do it in the long-term (since it doesn't put the updated `bandpass_array` anywhere), but it should be a quick patch that adds a bit of extra functionality we discussed in the Validation call.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,15 @@
 Changelog
 =========
 
+v2.3.0 [2022.01.19]
+===================
+
+Added
+-----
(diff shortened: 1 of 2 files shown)


steven-murray (approved):
This looks perfect, thanks @r-pascua.

r-pascua:
Awesome, thanks for the quick review! I'll handle merging and tagging after the tests complete.

PR #209, opened 2022-01-19
fix: set normalize_beams in ModelData.from_config

r-pascua:
It's now possible to set the `normalize_beams` parameter using the `ModelData.from_config` class method.


steven-murray (approved):
This looks right to me -- probably should be an option in the config itself in the future!

PR #212, opened 2022-02-18
fix: only add Trx bias to autocorr noise

r-pascua:
This PR makes sure that autocorrelations only receive a receiver temperature bias when using the `ThermalNoise` class to simulate noise. This is exactly what we did for H1C IDR2 validation, but I forgot to transfer it over to `hera_sim`.

Edit: Many tests broke as a result of recent changes to `vis_cpu`. I've also fixed those tests in this PR.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,14 @@
 Changelog
 =========
 
+v2.3.2 [2022.02.18]
+===================
+
+Fixed
+-----
+- The ``ThermalNoise`` class ``__call__`` method is now aware of baseline
+  length so that autocorrelations only receive a receiver temperature bias.
+
 v2.3.1 [2022.01.19]
 ===================
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,10 +5,20 @@ Changelog
 v2.3.2 [2022.02.18]
 ===================
 
+Added
+-----
+- ``_extract_kwargs`` attribute added to the ``SimulationComponent`` class. This
+  attribute is used by the ``Simulator`` to determine which optional parameters
+  should actually be extracted from the data.
+- ``antpair`` optional parameter added to the ``ThermalNoise`` class. This is
+  used to determine whether to simulate noise via the radiometer equation (as is
+  appropriate for a cross-correlation) or to just add a bias from the receiver
+  temperature (which is our proxy for what should happen to an auto-correlation).
+
 Fixed
 -----
-- The ``ThermalNoise`` class ``__call__`` method is now aware of baseline
-  length so that autocorrelations only receive a receiver temperature bias.
+- The ``Simulator`` class now correctly uses the auto-correlations to simulate
+  noise for the cross-correlations.
 
 v2.3.1 [2022.01.19]
 ===================
--- a/hera_sim/components.py
+++ b/hera_sim/components.py
@@ -39,6 +39,9 @@ class SimulationComponent(metaclass=ABCMeta):
 
     _alias: Tuple[str] = tuple()
 
+    # Keyword arguments for the Simulator to extract from the data
+    _extract_kwargs = set()
+
     def __init_subclass__(cls, is_abstract: bool = False):
         """Provide some useful augmentations to subclasses.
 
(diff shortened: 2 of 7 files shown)


r-pascua:
Posting an update to detail the problem and describe the solution I implemented.

It was found that the `Simulator` was not properly simulating noise, and this was revealed by comparing plots of the noise estimated from the data to the noise expected from the autos. It turns out that the issue was that the `Simulator` wasn't actually setting the `autovis` parameter to anything, since the `_initialize_args_from_model` method wasn't tracking this parameter. The `_initialize_args_from_model` method used to only track arguments without default values, and so it would entirely miss the `autovis` parameter in the `ThermalNoise` model, since it's optional. I've updated the logic of the `_initialize_args_from_model` method to also include parameters that are specified in a new class attribute common to all `SimulationComponent` subclasses that I've named `_extract_kwargs`. In practice, this is just something of a hotfix for the issue that maintains backwards compatibility with the `ThermalNoise` class and does not introduce any changes to the API. The idea is that this is a prototype for a feature @steven-murray and I have discussed on multiple occasions, where the model classes inform the `Simulator` of which parameters it should pull from the data. I say that it is a prototype because I think that the proper implementation should not only tell the `Simulator` which of the model arguments should be extracted from the data, but it should also tell the `Simulator` _how_ to extract those values, to the extent possible. (Of course, this stance is completely up for debate. It may turn out that the simplest thing to do is to just update the logic in the `Simulator._update_args` method whenever a new model class is introduced that adds a unique entry to the set of all `_extract_kwargs` entries.)

Anyway, this turned out to be a somewhat deeper and more nuanced issue than I originally thought it was, so I would like @steven-murray to take a look, as well as @jsdillon.

jsdillon:
As proof that this is working better, here's some plots for data simulated with this code:

BEFORE:

Screen Shot 2022-02-18 at 11 45 37 AM


AFTER:

image

aewallwi:
While we are on the topic of autocorrelations, I noticed that `gen_white_noise` method, which generates a gaussian random field,  is used to get noise fluctuations regardless of whether we are dealing with cross or autocorrelations. This method draws a complex normal number.

https://github.com/HERA-Team/hera_sim/blob/9576e78bd8807473c40152f9b6b61745ca9dc879/hera_sim/utils.py#L411

The thermal noise on autocorrelation should be drawn from a Gamma distribution. Is this done in `hera_sim`? If not, this should be corrected.

aewallwi:
> While we are on the topic of autocorrelations, I noticed that `gen_white_noise` method, which generates gaussian noise, is used to get noise fluctuations regardless of whether we are dealing with cross or autocorrelations.
> 
> https://github.com/HERA-Team/hera_sim/blob/9576e78bd8807473c40152f9b6b61745ca9dc879/hera_sim/utils.py#L411
> 
> The thermal noise on autocorrelation should technically be drawn from a Gamma distribution rather than a normal distribution. Is this done in `hera_sim`? If not, this should be corrected.
> 
> image

Also, maybe this converges to a Gaussian through the central limit theorem for large number of accumulations but maybe check / note whether this is actually the case.

jsdillon:
@aewallwi I think you want to post that stuff over on #213. I'd like to separate out the question of getting <V_ii> right with the question of getting the variance of V_ii right. One is key to H1C IDR3 validation, the other is useful more broadly.

r-pascua:
Thanks for the review and comments, everyone. I'm realizing that the level of generality here is coming across as more of a detriment on the development side, so I think it's worthwhile to see if we can figure out a better solution for v3. I think I'd like to open an issue or discussion on this topic and try to solicit community feedback, since I don't know really know what people want out of the code or what typical use cases outside are outside of the Validation stuff I do with it.

I'll leave the PR open for a few more hours for potential responses to this comment.

steven-murray:
@r-pascua I think an issue would be a good idea. Might be useful to tag in the issue where we previously discussed this. It's not absolutely clear to me that the current framework is not explicit enough -- that might be my own unfamiliarity with the code rather than any deficiency in the code itself. But I think its worthwhile revisiting before a new major version.

aewallwi:
> @aewallwi I think you want to post that stuff over on #213. I'd like to separate out the question of getting <V_ii> right with the question of getting the variance of V_ii right. One is key to H1C IDR3 validation, the other is useful more broadly.

I've copied this comment over to #213 and removed this one

r-pascua:
Issue about generality and complexity is posted (#214). Merging and closing this PR now.

PR #215, opened 2022-02-22
Fix interp

r-pascua:
This PR adds support for interpolating in time when there is a phase wrap in the LSTs. This works on the assumption that the time array is strictly increasing. It's possible that I may have over-simplified it and there might be poorly-handled edge cases, but I think this update does the right thing.

### DIFF
--- a/hera_sim/adjustment.py
+++ b/hera_sim/adjustment.py
@@ -482,12 +482,8 @@ def interpolate_to_reference(
             except TypeError:
                 raise TypeError("reference must be convertible to a UVData object.")
 
-        ref_time_to_lst_map = {
-            ref_time: ref_lst
-            for ref_time, ref_lst in zip(reference.time_array, reference.lst_array)
-        }
-        ref_times = np.array(list(ref_time_to_lst_map.keys()))
-        ref_lsts = np.array(list(ref_time_to_lst_map.values()))
+        ref_times, inds = np.unique(reference.time_array, return_index=True)
+        ref_lsts = reference.lst_array[inds].copy()
         ref_freqs = np.unique(reference.freq_array)
     else:
         if axis in ("time", "both"):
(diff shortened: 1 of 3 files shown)


hera_sim/adjustment.py
```diff
@@ -506,43 +504,30 @@ def interpolate_to_reference(
...
+            target_lsts[target_lsts < target_lsts[0]] += 2 * np.pi
+        if iswrapped(ref_lsts):
+            ref_lsts[ref_lsts < ref_lsts[0]] += 2 * np.pi
+
+        if np.any((ref_lsts < target_lsts[0]) | (ref_lsts > target_lsts[-1])):
             warn("Reference LSTs not a subset of target LSTs; clipping.")
-            key = np.argwhere(
-                np.logical_and(target_lsts[0] <= ref_lsts, ref_lsts <= target_lsts[-1])
-            ).flatten()
+            key = (target_lsts[0] <= ref_lsts) & (ref_lsts <= target_lsts[-1])
```
jsdillon, line 524: I don't love calling this `key` since it's not a dictionary key but more like a boolean slicing operation. I'd call it something more descriptive like `is_in_range`

hera_sim/adjustment.py
```diff
@@ -506,43 +504,30 @@ def interpolate_to_reference(
...
-            np.logical_or(ref_freqs < target_freqs[0], ref_freqs > target_freqs[-1])
-        ):
+        if np.any((ref_freqs < target_freqs[0]) | (ref_freqs > target_freqs[-1])):
             warn("Reference frequencies not a subset of target frequencies; clipping.")
-            key = np.argwhere(
-                np.logical_and(
-                    target_freqs[0] <= ref_freqs, ref_freqs <= target_freqs[-1]
-                )
-            ).flatten()
+            key = (target_freqs[0] <= ref_freqs) & (ref_freqs <= target_freqs[-1])
```
jsdillon, line 530: Same comment.

jsdillon (approved):
Looks good to me! Just two nomenclature tweaks.

jsdillon (approved):

Issue #211, opened 2022-02-04
Check UVData.future_array_shapes when caching ModelData.freqs 


piyanatk:
As in the title.
https://github.com/HERA-Team/hera_sim/blob/9576e78bd8807473c40152f9b6b61745ca9dc879/hera_sim/visibilities/simulators.py#L193-L196
These lines assume that `current_array_shapes` is being used. If the UVData uses `future_array_shape`, it will break SkyModel evaluation (and maybe at other places.)
https://github.com/HERA-Team/hera_sim/blob/9576e78bd8807473c40152f9b6b61745ca9dc879/hera_sim/visibilities/simulators.py#L83

r-pascua:
Thanks for making an issue about this. I don't necessarily know if I'd classify this as a bug, but this is something that has been in the back of my mind for a few months now. I chatted with Paul and was informed that the community would receive sufficient notice about when the future array shapes will become the new API, so I decided to put off dealing with it. There are also at least a few places in the `Simulator` class that need to be updated to accommodate using the future array shape.

I think I might want to make this a milestone for version 3, which I'm envisioning will implement the merger of the `Simulator` class and the `VisibilitySimulation` class. Curious to hear thoughts from @steven-murray.

piyanatk:
It looks like `initialize_uvdata_from_params` is default to use future array shape from `pyuvdata=>2.2.6`. I do not see this mentioned in the pyuvdata changelogs though. Maybe @bhazelton can verify this.

PR #216, opened 2022-02-22
Temporary fix for future array shape

piyanatk:
pyuvdata switched to using `future_array_shapes` by default from `pyuvdata>=2.7.0`. Many functionalities of `hera_sim=2.x.x` relies on the `current_array_shapes` (see, e.g., #211). This PR provides a temporary fix to this issue  The full fix is planned for `hera_sim=3.x.x`.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,13 @@
 Changelog
 =========
 
+dev-version
+===========
+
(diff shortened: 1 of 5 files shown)


steven-murray (approved):
Thanks @piyanatk this looks correct

PR #218, opened 2022-02-24
fix: inferred integration time in ThermalNoise

r-pascua:
Without this fix, the inferred integration time for a phase-wrapped LST array is negative for most use cases, which causes the calculated noise to take on nan values. This commit fixes that issue, but it also adds an exception for the case where the LSTs are phase-wrapped and the sky temperature is intended to be calculated from an interpolation object. (This case would have been handled poorly prior to this fix, so this just makes the error more explicit.)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,6 +5,15 @@ Changelog
 dev-version
 ===========
 
+Added
+-----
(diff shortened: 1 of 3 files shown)


jsdillon (approved):
Looks good to me!

Issue #219, opened 2022-03-15
URLs starting with "git://" in setup.cfg need to be changed to "https://"


bhazelton:
Github no longer supports these urls so pip fails trying to install hera_sim.

See https://github.blog/2021-09-01-improving-git-protocol-security-github/

bhazelton:
I made this because I'm seeing an error in pyuvsim testing, but I just realized that testing is actually not breaking because of this (although this is still an issue) because we use the `--no-deps` flag when we pip install hera_sim. The problem is actually lower down in the HERA stack, in uvtools and probably other places.

bhazelton:
Update: the hera_cal issue has now been fixed, so our test on pyuvsim works. But I think these URLs should be changed.

philbull:
Fixed in #220

PR #220, opened 2022-03-22
Add dummy select() method to PolyBeam classes

philbull:
Trivial PR to add a dummy `select()` method to the `PolyBeam` class, so it can be used by the new-style hera_sim `VisibilitySimulator` class.

Incidentally, this also fixes an issue with the pre-commit hooks using a now-deprecated GitHub access URL.


philbull:
@steven-murray Hopefully fixed #219 now...

steven-murray (approved):
Thanks @philbull this looks good. Looks like one of our deps is depending on a `http://` url for `hera_qm` which is failing on one OS, but that's not our problem directly.

philbull:
Thanks, I'll merge now if that's OK. (I think the test will pass, it's just giving a HTTP request rate error)

PR #222, opened 2022-04-20
fix: warn about using use_pixel_beams

steven-murray:
This sets the default value of `use_pixel_beams` to False, because it's a bad idea. 

Since it changes the API (in the sense of a default change), but I didn't want to make it a major-version thing, I emit a warning when the default value is used.


r-pascua (approved):
This looks good to me. I agree that this it feels silly to make a new major version release just for changing a default parameter, and that raising a warning is a nice middle ground. Thanks!

PR #225, opened 2022-05-26
fix: signature of polybeam.interp

steven-murray:
Fixes the signature of `PolyBeam.interp`. Should fix the issues seen in https://github.com/RadioAstronomySoftwareGroup/pyuvsim/pull/404


bhazelton (approved):
Looks good, thanks for the fast fix!

PR #226, opened 2022-06-07
maint: remove all git repo deps

steven-murray:
This removes all the git-repo dependencies from the setup.cfg.

Most of these were easy, since they now have PyPI packages. The tricky ones were hera_gpu and healvis. For hera_gpu, I think we don't actually need it after all -- I replaced it with the underlying required packages. For healvis, it was only really needed for tests (it's an optional dependency). I've now removed it from being installed as an optional dependency (it can still be installed manually, and the module used, but it has been marked as officially deprecated). We still install it for CI tests until version 4, but following that it can be permanently removed.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,15 +5,20 @@ Changelog
 dev-version
 ===========
 
+v2.3.4 [2022.06.08]
+===================
+
 Added
 -----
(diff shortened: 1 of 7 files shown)


bhazelton (approved):
Looks good to me!

r-pascua (approved):
Looks good to me!

PR #227, opened 2022-06-09
maint: add auto-upload to pypi in CI

steven-murray:
Should auto-deploy on tag creation now!

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -9,10 +9,6 @@ jobs:
   build-n-publish:
     name: Build and publish to PyPI
     runs-on: ubuntu-latest
-    defaults:
-      run:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yaml
```diff
@@ -0,0 +1,41 @@
...
+      - '*'
+
+jobs:
+  build-n-publish:
+    name: Build and publish to PyPI
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        # Adding -l {0} helps ensure conda can be found properly.
+        shell: bash -l {0}
```
mkolopanis: I don't think we need the login shell
```suggestion

```

mkolopanis (approved):

PR #229, opened 2022-06-30
add check for clobber keyword in obsparams file

jburba:
I modified the `hera-sim-vis.py` script to include a check for the clobber keyword in the filing parameters of the obsparam file.  If the clobber keyword is not present in the obsparams file, then clobber is set to False.  Otherwise, clobber is set to the value in the obsparam file.

### DIFF
--- a/scripts/hera-sim-vis.py
+++ b/scripts/hera-sim-vis.py
@@ -165,10 +165,7 @@ if __name__ == "__main__":
         outfile = (
             base_path / f"{cfg_filing['outfile_name']}.{cfg_filing['output_format']}"
         )
-        if "clobber" not in cfg_filing:
-            clobber = False
-        else:
-            clobber = cfg_filing["clobber"]
+        clobber = cfg_filing.get("clobber", False)
 
         # Write output
         cns.print("Writing output... ", end="")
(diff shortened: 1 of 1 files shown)


scripts/hera-sim-vis.py
```diff
@@ -165,10 +165,14 @@ def cprint(*args, **kwargs):
         outfile = (
             base_path / f"{cfg_filing['outfile_name']}.{cfg_filing['output_format']}"
         )
+        if "clobber" not in cfg_filing:
+            clobber = False
```
piyanatk, line 169: Maybe also print a warning here that `clobber` is being set to `False`.

piyanatk (approved):
This looks good @jburba. Just one suggestion but I approved.

scripts/hera-sim-vis.py
steven-murray: I don't think we'll need the warning. If clobber is False, and the file already exists, then an error will be raised later anyway, right? I think setting to False is non-destructive and so doesn't need a warning.

steven-murray: To make the code slightly simpler, could remove all this if/else, and just replace the line 171 with `clobber=cfg_filing.get('clobber', False)`

jburba: Yes, pyuvdata raises an error if the file exists and clobber is set to False.  And I forgot about the get function.  I'll make that change.

PR #233, opened 2022-08-10
fix: update VisCPU simulator wrapper to match vis_cpu v1

steven-murray:
The `use_pixel_beams` and `bm_pix` options are no longer available. `use_pixel_beam` is always effectively False. This has been deprecated already for a while.

One upshot of this is that some of the beam tests no longer work -- some of them were comparing a pixel beam sim to a non-pixel-beam sim. I don't think this was a particularly good strategy for testing the beams anyway, and I've removed the code. But we should probably come up with something to replace those kinds of tests.

### DIFF
--- a/hera_sim/__init__.py
+++ b/hera_sim/__init__.py
@@ -10,7 +10,7 @@ try:
     DATA_PATH = Path(__file__).parent / "data"
     CONFIG_PATH = Path(__file__).parent / "config"
     __version__ = version(__name__)
-except PackageNotFoundError:
+except PackageNotFoundError:  # pragma: no cover
     # package is not installed
     pass
 
--- a/hera_sim/__init__.py
+++ b/hera_sim/__init__.py
@@ -2,9 +2,9 @@
 from pathlib import Path
 
 try:
-    from importlib.metadata import version, PackageNotFoundError
+    from importlib.metadata import PackageNotFoundError, version
 except ImportError:
-    from importlib_metadata import version, PackageNotFoundError
+    from importlib_metadata import PackageNotFoundError, version
 
 try:
     DATA_PATH = Path(__file__).parent / "data"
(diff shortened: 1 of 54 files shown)


hera_sim/__init__.py
```diff
@@ -2,36 +2,43 @@
 from pathlib import Path
 
 try:
-    from importlib.metadata import version, PackageNotFoundError
+    from importlib.metadata import PackageNotFoundError, version
 except ImportError:
```
r-pascua, line 6: Maybe want to add a `# pragma: nocover` here so codecov ignores this?

hera_sim/components.py
```diff
@@ -37,7 +38,7 @@ class determine how to apply the effect simulated by
     #: Whether this systematic multiplies existing visibilities
     is_multiplicative: bool = False
 
-    _alias: Tuple[str] = tuple()
+    _alias: tuple[str] = tuple()
```
r-pascua, line 41: Does the change from `Tuple` to `tuple` in the typing have any effect? It looks like one of the new style-enforcing features you added changed `Tuple` to `tuple` and `Dict` to `dict` wherever those occur.

hera_sim/tests/test_beams.py
```diff
@@ -252,38 +250,38 @@ def test_rotations(self, antennas, sources):
         rvals = np.linspace(0.0, 180.0, 31, dtype=int)
 
         rotations = np.zeros(rvals.size)
-        pix_results = np.zeros(rvals.size)
+        # pix_results = np.zeros(rvals.size)
```
r-pascua, line 253: Do we want to remove this instead of just comment it out?

hera_sim/tests/test_beams.py
```diff
@@ -252,38 +250,38 @@ def test_rotations(self, antennas, sources):
...
 
         # Check that the maximum difference between pixel beams/direct calculation
         # cases is no more than 5%. This shows the direct calculation of the beam
         # tracks the pixel beam interpolation. They won't be exactly the same.
-        np.testing.assert_allclose(pix_results, calc_results, rtol=0.05)
+        # np.testing.assert_allclose(pix_results, calc_results, rtol=0.05)
 
         # Check that rotations 0 and 180 produce the same values.
-        assert pix_results[0] == pytest.approx(pix_results[-1], abs=1e-8)
+        # assert pix_results[0] == pytest.approx(pix_results[-1], abs=1e-8)
```
r-pascua, line 272: Here too.

r-pascua (approved):
From what I can tell, this all looks good! When you get a chance, maybe delete the commented out stuff in the tests. On that note, it's probably worthwhile for us to brainstorm a bit about what the purpose of the beam-related tests was and whether it's appropriate to be testing that here (i.e. whether we should bother rethinking and rewriting the tests).

Thanks for doing this!

P.S. In the future, if we're adding some new cool features (like `isort`), then can we keep that to a single PR? Having the whole code base be reformatted is pretty noisy and makes it a little difficult to figure out what was just reformatting and what was genuinely changed as part of the PR.

hera_sim/__init__.py
steven-murray: I've added a global ignore of lines with `except ImportError` in the .coveragerc

hera_sim/components.py
steven-murray: No, no effect, just one less import to make. You need I think py38 to use it.

steven-murray:
Thanks @r-pascua! Good call on all the commented code -- I've removed it. 

And yeah, sorry about the isort thing. That was a brain fart.

PR #234, opened 2022-08-15
Profiling additions

steven-murray:
Adds profiling option to the `hera-sim-vis.py` script, which makes it possible to do line-profiling as part of the run.

### DIFF
--- a/scripts/hera-sim-vis.py
+++ b/scripts/hera-sim-vis.py
@@ -78,7 +78,16 @@ if __name__ == "__main__":
         help="If given, do line-profiling on the simulation, and output to given file.",
     )
     parser.add_argument(
-        "-p", "--extra-profile-func", type=str, action="append", dest="profile_funcs"
+        "-p",
+        "--extra-profile-func",
+        type=str,
+        action="append",
+        dest="profile_funcs",
+        help=(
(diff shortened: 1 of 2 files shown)


scripts/hera-sim-vis.py
```diff
@@ -132,6 +136,16 @@ def cprint(*args, **kwargs):
         for fnc in simulator._functions_to_profile:
             profiler.add_function(fnc)
 
+        # Now add any user-defined functions that they want to be profiled.
+        # Functions must be sent in as "path.to.module:function_name" or
+        # "path.to.module:Class.method".
```
r-pascua, line 141: Can you add the part about the required format to the `help` info for the `profile_funcs` argument?

r-pascua (changes requested):
Looks good to me! Just expose the required format for the `profile_funcs` to the user and I'll approve. (We'll also want to make a note about this in any tutorial that touches on the profiling capabilities.)

scripts/hera-sim-vis.py
steven-murray: Ah yeah, good idea.

r-pascua (approved):
Thanks for making the change! It all looks good to me!

PR #237, opened 2022-09-07
Memory profiling

### LINKED ISSUE #191: Bug in UVSim beam_dict parsing
`ModelData._process_beam_ids` returns a dictionary, not an array of integer as indicated in the typing.

https://github.com/HERA-Team/hera_sim/blob/153eef4cb1d0b38666791a116187f106ab115fb2/hera_sim/visibilities/simulators.py#L116-L143

This causes an error in `UVSim.simulate`, which expects an integer array of beam_ids.
https://github.com/HERA-Team/hera_sim/blob/153eef4cb1d0b38666791a116187f106ab115fb2/hera_sim/visibilities/pyuvsim_wrapper.py#L20-L25

`VisCPU` is currently working okay (with the dictionary `beam_ids` passed from `ModeData`).

steven-murray:
This PR adds quite a lot of functionality/performance improvement.

Fixes #191 

Highlights include:

- New `LogRender` and `RichHandler` classes in `cli_utils.py`. These are extensions of the `RichHandler` from `rich` itself, that add a column for current RAM usage, and also give the option to print out the time since start of script, instead of current absolute time.
- New basic tests of `hera-sim-vis.py`.
- New `_blt_order_kws` class attribute for the `VisibilitySimulator` subclasses, which specify how to reorder the blt axis of the constructed simulation (if required) before simulating. This is useful as it allow the data to NOT be reordered (which can take some time). 
- Many more logging statements throughout the code to enable easier performance profiling. THese statements don't show by default (they are typically INFO level).
- Turn off `run_check_acceptability` when constructing from obsparams (it significantly increases peak memory usage, and is unnecessary since we just made the object from scratch).
- New (optional) `compress_data_model()` method on `VisibilitySimulator` subclasses that allows unnecessary metadata in the UVData object to be dropped before simulation (usually this will have to be restored afterwards with the associated `restore_data_model()` method). This is just to reduce peak memory usage.
- New `check_antenna_conjugation` parameter for the `VisCPU` simulator -- allow turning off the check for antenna conjugation. Usually, doing this is safe, but when constructing the object from obsparams, it really doesn't make sense to do this time-consuming check.
- In `VisCPU`, ability to not copy the entire data array, but rather just insert data into the existing one. This only triggers if the whole array is made of zeros (i.e. we wouldn't be over-riding data).
- A much much faster `_reorder_vis` for when the `blt_order` of the UVData object is `('time', 'ant1')`. In this case, the output from vis_cpu is already ordered correctly, we just have to take parts of it and stick it in the `data_array`. This reduces reordering time by like 99%.
- Also in `_reorder_vis`, I swapped the loop ordering (for when `blt_order` is not so nice). This halved the time for the reordering.
- Added printing of simulation configuration in `hera-sim-vis.py`
- The `--compress` option to `hera-sim-vis.py` is no longer a boolean flag, but takes a file argument. This file will be written as a cache of the baseline-time indices required to keep when compressing by redundancy. If the file exists, the code will use that cache instead of regenerating the indices, which can itself take a lot of time. This saves many many minutes on H4C-style sims.
- Add a `--log-level` option to `hera-sim-vis.py` to change the verbosity.
- Added a `--dry-run` option to `hera-sim-vis.py` to construct the sim config and report on it, but not do the actual sim. Useful for testing setup time and making sure the setup looks correct.

Overall, with these improvements, I was able to take a run of 5760 times and 126 antennas from 5min + >30min (for simulation + setup/reorder/write) down to 5min + ~40sec.


steven-murray:
I should say that this is waiting on https://github.com/RadioAstronomySoftwareGroup/pyuvdata/pull/1205 and https://github.com/RadioAstronomySoftwareGroup/pyuvsim/pull/410 before it will pass tests.

piyanatk:
This line in `hera-sim-vis.py` will error when using analytic beams
https://github.com/HERA-Team/hera_sim/blob/03d13da0b921216855e10ac15b36d83013631254/scripts/hera-sim-vis.py#L128
Should I open an issue? It is still in a PR so I am not sure if I should

hera_sim/visibilities/cli.py
```diff
@@ -0,0 +1,341 @@
...
+    #             _fnc = getattr(_fnc, att)
+    #         profiler.add_function(_fnc)
+
+    # else:
+    #     profiler = None
+
+    # if args.profile:
+    #     profiler.runcall(main, args, profiler, simulator)
+    # else:
+    #     main(args, profiler, simulator)
```
piyanatk, line 341: We should probably tidying this up

setup.cfg
```diff
@@ -29,6 +29,7 @@ install_requires =
     astropy-healpix
     cached-property
     deprecation
+    hera-cli-utils>=0.1.0
     numpy>=1.18
     pyuvdata
```
piyanatk, line 34: We should pin `pyuvdata` at least to the version that require the new frequency array shape, maybe even the more recent version, I am not sure. This is implicitly enforced from `pyuvsim` requirement, I think, but being explicit is better.

piyanatk (approved):
Look good @steven-murray. Just some tidy up.

steven-murray:
@piyanatk thanks for the review! I've addressed your two comments.

setup.cfg
steven-murray: Pinned to 2.2.12

hera_sim/visibilities/cli.py
steven-murray: Removed, thanks for spotting!

PR #246, opened 2022-12-06
feat: mutual coupling a la Josaitis+22

r-pascua:
This PR implements the "first order coupling" model described in [Josaitis et al. 2022](https://academic.oup.com/mnras/article/514/2/1804/6564718) via the `sigchain.MutualCoupling` class. Some additional utility functions were also added to help speed things up. This *should* support fully or partially (linearly) polarized data, although the implementation is a little more memory hungry than it needs to be for unpolarized data (this is a design choice I made arbitrarily&mdash;the code is much neater this way).

At a high level, the procedure goes like this:
1. Calculate a large "coupling matrix"
2. Reshape the input data array to a shape compatible with the coupling matrix
3. Calculate the first-order coupling (what I'll call the crosstalk visibilities) via `xt_vis = data @ coupling.H + coupling @ vis`
4. Reshape the crosstalk visibilities to comply with the input data array and return the result

This isn't currently ready to be merged; I still need to add documentation and a few unit tests. I'm a bit hesitant to try optimizing further&mdash;the only things I can really think of to improve performance require writing some wrapped C-code, or figure out how to access the BLAS/LAPACK functions that `numba` uses under the hood (assuming I actually have an idea of how `numba` works...). Here are some specs for timing tests I did for a H4C-sized file with NRAO resources, using 8 cores (I don't think I had any other processes running, but I'm not sure if anything else was running on the same node at the time I ran the tests).

First, the data parameters:
* 2 integrations from JD 2459135
* 1510 channels (had to cut out things below 50 MHz)
* 103 antennas (5356 baselines, 933 redundant groups)
* 4 polarizations

Now for the various steps:
1. Calculate coupling: ~15 seconds
2. Reshape input data: ~2.5 seconds
3. Do matrix products and sum: ~6.5 seconds
4. Reshape crosstalk visibilities: ~1.3 seconds

Will add some plots at a later date, but preliminary simulated visibilities look good.


hera_sim/tests/test_sigchain.py
```diff
@@ -457,13 +460,144 @@ def test_mutual_coupling():
...
+    uvb.spw_array = np.zeros(1, dtype=int)
+
+    # Finally, all the shape parameters
+    uvb.Naxes1 = az.size
+    uvb.Naxes2 = za.size
+    uvb.Naxes_vec = 2
+    uvb.Ncomponents_vec = 2
+    uvb.Nfeeds = 2
+    uvb.Nfreqs = freqs.size
+    uvb.Nspws = 1
```
steven-murray, line 537: So, all this could essentially be achieved with a `AnalyticBeam.to_uvbeam()` method?

hera_sim/utils.py
```diff
@@ -488,3 +498,286 @@ def _listify(x):
...
+    # We don't have numba, so we need to do this a bit more slowly.
+    pol_slices = {"x": slice(None, None, 2), "y": slice(1, None, 2)}
+    for i, ai in enumerate(antenna_numbers):
+        for j, aj in enumerate(antenna_numbers[i:]):
+            j += i
+            for k, pol in enumerate(pol_array):
+                uvd_inds = np.argwhere(
+                    (ant_1_array == ai) & (ant_2_array == aj)
+                ).flatten()
+                p1, p2 = uvutils.polnum2str(pol)
```
steven-murray, line 582: In my experience, this function call actually takes a bit of time. You could cache the result into a dict outside the big loops.

steven-murray:
Thanks @r-pascua , this is excellent work. I would love to see a notebook demo with some plots that "show" what the tests test. If we could go through such a notebook on a validation call, I think that would be the best final review. Otherwise, the code changes look good to me. 

Funnily enough, I thought I saw such a notebook in the files changed in this PR, but now I can't see it...

hera_sim/tests/test_sigchain.py
r-pascua: Yes, though you'd need to provide an `az_array`, `za_array`, and `freq_array`, and then all the other stuff (like the `feed_name` and whatnot) could be optional parameters.

hera_sim/tests/test_sigchain.py
```diff
@@ -457,13 +460,144 @@ def test_mutual_coupling():
...
+        resistance=resistance,
+        reflection=reflection,
+    )
+    assert not np.any(np.isclose(data, sample_uvdata.data_array))
+
+
+def test_mutual_coupling_bad_ants(sample_uvdata, sample_coupling):
+    full_array = dict(zip(*sample_uvdata.get_ENU_antpos()[::-1]))
+    bad_array = {ant: full_array[ant] for ant in range(len(full_array) - 1)}
+    with pytest.raises(ValueError) as err:
```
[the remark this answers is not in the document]
r-pascua: Good catch! I realized this when I was looking through some other tests where I had used the `match=...` argument, but I didn't want to change it unless told otherwise.

r-pascua:
Thanks for the review! I do have a notebook in the works, but it was a little messy and I wasn't sure if it was necessary for this PR. I'll have it ready by next week's validation call.

hera_sim/utils.py
r-pascua: I'm not sure I understand what you mean&mdash;the big loops only go over unique antenna pairs, so there's nothing to gain from caching locally. Moving the `uvd_inds = ...` snippet to outside the polarization loop would save some time when dealing with fully-polarized data, so I suppose I could make that change.

The only other thing I can think of would only save time on the second reshape (i.e. reshaping the crosstalk to match the data shape) by adding something like `uvd_ind_cache={}` to the function signature, but I'm not sure how I feel about that. I suppose it's fine if we're pretty confident that we're only going to be reshaping data for one `UVData` object per session, but I think that implementing a cache like this would cause some tests to break. Another option would be to require the user to provide the cache (i.e. instead doing `uvd_ind_cache=None`, then adding a `if uvd_ind_cache is None: uvd_ind_cache = {}` clause to the function body), but manually providing the cache feels a little annoying. (Also note that both of these options would only work for the non-jitted version of the code.)

Curious to hear what you think.

hera_sim/tests/test_simulator.py
```diff
@@ -149,6 +149,30 @@ def __call__(self, lsts, freqs):
...
+    class TestBase:
+        pass
+
+    class Test(TestBase):
+        return_type = "full_array"
+        attrs_to_pull = dict(
+            freqs="freqs",
+            ant_1_array="ant_1_array",
+            pols="polarization_array",
+        )
```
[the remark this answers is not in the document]
r-pascua: Sure, but I think I'd need to think a little carefully to do it right. Are you thinking of doing that so we can omit things from `attrs_to_pull`? Like, instead of the above, we would just do `attrs_to_pull = {"pols": "polarization_array"}`?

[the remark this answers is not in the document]
r-pascua: OK, I've implemented it now. Relevant changes are in the `Simulator._initialize_args_from_model` method (in particular, lines 932-935).

hera_sim/utils.py
steven-murray: I actually just meant the `polnum2str` function. You could do something like 

```python
polnumstr = {pol: uvutils.polnum2str(pol) for pol in pol_array}
    for i, ai in enumerate(antenna_numbers):
        for j, aj in enumerate(antenna_numbers[i:]):
            j += i
            for k, pol in enumerate(pol_array):
                p1, p2 = polnumstr[pol]
```

In your case it might not make much difference... I can't remember if it was when including the `x_orientation` that it was slower. But I definitely found that calling this for all baselines in H6C was a bottleneck.

hera_sim/tests/test_sigchain.py
steven-murray: Yeah, I think this is in the works in pyuvdata, but for now this is good.

r-pascua: Oh, nice! That would be super convenient.

hera_sim/utils.py
r-pascua: Ohh, gotcha. I didn't think that'd be a bottleneck, but that's a nice solution. I'll make the change. Thanks!

hera_sim/simulate.py
```diff
@@ -129,44 +129,51 @@ def __init__(
...
         return np.unique(self.data.freq_array) / 1e9
 
-    @cached_property
+    @property
     def times(self):
         """Simulation times in JD."""
         return np.unique(self.data.time_array)
 
-    @cached_property
+    @property
```
steven-murray, line 176: What is the reason that this was necessary? Operations like `np.unique` are often non-negligible, which is why I cached them originally.

steven-murray (commented):
Thanks @r-pascua this is brilliant. I think everything is good to go now, other than one question.

hera_sim/simulate.py
r-pascua: Ah, this was so you could do a `select` on the `UVData` object and not have the wrong properties cached. It wasn't necessary for the features developed in this PR, but it was something I had realized as I was making some tweaks to the `Simulator`.

steven-murray: Hmm, I guess this is a thing that could happen. I guess having them as `@property` is better for now as it ensures correctness. But we might want to think harder if it becomes a performance problem later on.

r-pascua: Hmm. I don't think we'll see much of a difference from not caching these attributes, since the attributes that are determined via a `np.unique` call are only ever queried once per simulation component that's applied (they're pulled from the `Simulator` at the beginning of the `_iteratively_apply` method, before any loops are entered). We can keep an eye on it though.

If it does end up being a non-negligible slowdown, then I think the thing to do is probably make a `select` method for the `Simulator` class that performs the selection on the underlying `UVData` object and then deletes cached attributes that would've been affected by the selection.

steven-murray: Agreed. Well, green light from me!

steven-murray (approved):

PR #250, opened 2023-02-24
fix: make uvbeam optional in MutualCoupling init

r-pascua:
It isn't necessary to provide an argument for the `uvbeam` parameter when initializing the `MutualCoupling` class, but this parameter was required prior to this update. This PR fixes that issue.


steven-murray (approved):
Good catch, should've noticed this in the original PR, sorry!

PR #251, opened 2023-03-29
H4C Bandpass

r-pascua:
This PR adds a model for the H4C bandpass, as well as a tanh tapering function to assist in getting the right rolloff when adding a realistic level of per-antenna perturbations to the base bandpass. This PR also adds a tool for interpolating complex-valued reflection coefficients.


hera_sim/interpolators.py
```diff
@@ -360,3 +361,33 @@ def __init__(self, datafile, **interp_kwargs):
...
+
+class Reflection(FreqInterpolator):
+    """Complex reflection coefficient interpolator."""
+
+    def __init__(self, datafile, **interp_kwargs):
+        if "interpolator" not in interp_kwargs:
+            interp_kwargs["interpolator"] = "interp1d"
+        super().__init__(datafile, **interp_kwargs)
+        self._obj = "reflection"
+        self._check_format()
```
steven-murray, line 374: Why isn't this called in the `__init__()` of the base class?

hera_sim/interpolators.py
```diff
@@ -360,3 +361,33 @@ def __init__(self, datafile, **interp_kwargs):
...
         self._obj = "bandpass"
         self._check_format()
+
+
+class Reflection(FreqInterpolator):
+    """Complex reflection coefficient interpolator."""
+
+    def __init__(self, datafile, **interp_kwargs):
+        if "interpolator" not in interp_kwargs:
+            interp_kwargs["interpolator"] = "interp1d"
```
steven-murray, line 371: not sure what the point of this is -- you have hardcoded `interp1d` as the interpolator in the functions, so there's no point in setting `interpolator`?

hera_sim/utils.py
```diff
@@ -678,6 +678,24 @@ def find_baseline_orientations(
     return antpair2angle
 
 
+def tanh_window(x, x_min=None, x_max=None, scale_low=1, scale_high=1):
+    if x_min is None and x_max is None:
```
steven-murray, line 682: Seems like the warning message could include the statement that one should provide either `x_min` or `x_max`

steven-murray (commented):
Thanks @r-pascua, this is near-perfect. I just had a couple of simple comments.

hera_sim/utils.py
r-pascua: Good catch, I'll fix it. Thanks!

hera_sim/interpolators.py
r-pascua: Oh, the code was breaking before this. The issue is that when the class is initialized, it tries to read the file assuming that it contains polynomial coefficients unless the `interpolator` is set to something else. So this is effectively setting the default `interpolator` to be `interp1d` unless specified otherwise. It's hacky for sure, but it's a consequence of me not having enough foresight when I initially wrote this module.

hera_sim/interpolators.py
r-pascua: It should be, and I can make that change. Somewhere down the line I want to rewrite this entire module, though, since I didn't fully understand the scope of the things we'd typically want to interface with when I wrote this module.

steven-murray: Story of our life 😆

hera_sim/interpolators.py
steven-murray: Ahhh I see. No problem then.

steven-murray (approved):
Thanks @r-pascua, this looks good to me. The only changes I'm going to request are an entry in the changelog, and to check what's going wrong in the RTD build. It may be that I've fixed that in the #237 PR, so we could also try waiting til that's merged (just polishing it off now).

PR #254, opened 2023-04-12
ci: pin numpy in ci due to numba

steven-murray:
pyuvsim external tests of herasim were breaking, and it seems it's because the latest numpy (1.24) and some versions of numba are not compatible (see https://github.com/numba/numba/issues/8615). This pins numpy<1.24 until that problem is sorted out.


steven-murray:
@r-pascua I think this is ready to go. I'm not sure why the coverage is not picking up the calling of the `jit_reshape_vis_xxx` functions, because I know they are being run :/

CHANGELOG.rst
```diff
@@ -5,6 +5,13 @@ Changelog
 dev-version
 ===========
 
+Breaking Changes
```
r-pascua, line 8: Should this be added to a `v4.0.0 [yyyy.mm.dd]` entry?

CHANGELOG.rst
```diff
@@ -5,6 +5,13 @@ Changelog
...
+----------------
+- Removed the ``HealVis`` wrapper. Use ``pyuvsim`` instead.
+Changed
+-------
+- Updated ``PolyBeam`` to use future array shapes.
```
r-pascua, line 13: I'd maybe also add that the package as a whole has been updated to use future array shapes.

hera_sim/adjustment.py
```diff
@@ -144,6 +144,8 @@ def adjust_to_reference(
     # Check if the target object is a Simulator, but work with a UVData object.
     target_is_simulator = isinstance(target, Simulator)
     target = _to_uvdata(target)
+    if not target.future_array_shapes:
```
r-pascua, line 147: I think we can add `# pragma: nocover` here.

hera_sim/simulate.py
```diff
@@ -892,6 +893,9 @@ def _initialize_data(
                 "Otherwise, keywords must be provided to build a UVData object."
             )
 
+        if not self.data.future_array_shapes:
```
r-pascua, line 896: I think we can add `# pragma: nocover` here as well.

hera_sim/utils.py
```diff
@@ -552,23 +552,25 @@ def reshape_vis(
...
     """
     if invert:
-        out = np.zeros((ant_1_array.size, 1, n_freqs, n_pols), dtype=complex)
+        out = np.zeros((ant_1_array.size, n_freqs, n_pols), dtype=complex)
     else:
         out = np.zeros((n_times, n_freqs, 2 * n_ants, 2 * n_ants), dtype=complex)
 
     # If we have numba, then this is a bit faster.
     if HAVE_NUMBA and use_numba:
-        return jit_reshape_vis(
```
r-pascua, line 561: Maybe just put add `# pragma: nocover` after `if HAVE_NUMBA and use_numba:` since coverage seems to be missing this. (I've found that coverage generally has a hard time with anything that has been jitted, although it's weird that it doesn't register calling these functions.)

r-pascua (approved):
I think this looks good to go! I'd be happy to have this merged once the comments are addressed.

CHANGELOG.rst
steven-murray: Yeah, I was thinking of doing that in a separate PR, but might as well do it here.

CHANGELOG.rst
steven-murray: Done.

hera_sim/adjustment.py
steven-murray: yeah, good point. We'll be removing it anyway soon.

hera_sim/simulate.py
steven-murray: 👍🏻

PR #257, opened 2023-05-24
fix: pass spline_interp_opts

piyanatk:
This PR fix an issue where the `spline_interp_opts` keyword does not get passed to the `vis_cpu` wrapper, rendering linear vs cubic interpolation indifferent


steven-murray (approved):
This looks right to me. There's an error in the tests, but I can deal with it in the `memory-profiling` branch.

PR #263, opened 2023-07-06
fix: call to run_with_profiling

steven-murray:
This fixes the call to `run_with_profiling` in `hera-sim-vis.py`.


piyanatk (approved):
Oh. Haha. This is super simple. I tried to read through `hera_cli` code but couldn't figure out what was wrong ... maybe I was sleepy. Thanks for the fix @steven-murray .

PR #266, opened 2023-08-01
Fix-for-pyuvdat2.4

steven-murray:
Updates API calls for pyuvdata to be able to use v2.4.0.

### DIFF
--- a/hera_sim/sigchain.py
+++ b/hera_sim/sigchain.py
@@ -1086,12 +1086,15 @@ class MutualCoupling(Crosstalk):
             else:
                 power_beam = uvbeam.copy()
                 power_beam.efield_to_power()
-                if power_beam.interpolation_function is None:
-                    power_beam.interpolation_function = pixel_interp
+                # if power_beam.interpolation_function is None:
+                #    power_beam.interpolation_function = pixel_interp
                 if power_beam.freq_interp_kind is None:
                     power_beam.freq_interp_kind = freq_interp
                 power_beam = power_beam.interp(
-                    freq_array=freqs * units.GHz.to("Hz"), new_object=True
+                    freq_array=freqs * units.GHz.to("Hz"),
+                    new_object=True,
+                    interpolation_function=pixel_interp,
+                    #                    freq_interp_func=freq_interp,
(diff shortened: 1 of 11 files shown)


hera_sim/sigchain.py
```diff
@@ -1086,12 +1086,15 @@ def build_coupling_matrix(
             else:
                 power_beam = uvbeam.copy()
                 power_beam.efield_to_power()
-                if power_beam.interpolation_function is None:
-                    power_beam.interpolation_function = pixel_interp
+                # if power_beam.interpolation_function is None:
+                #    power_beam.interpolation_function = pixel_interp
```
r-pascua, line 1090: It looks like this is now taken care of in the call to `interp`. Shouldn't these lines be removed instead of commented out?

hera_sim/sigchain.py
```diff
@@ -1086,12 +1086,15 @@ def build_coupling_matrix(
...
                 if power_beam.freq_interp_kind is None:
                     power_beam.freq_interp_kind = freq_interp
                 power_beam = power_beam.interp(
-                    freq_array=freqs * units.GHz.to("Hz"), new_object=True
+                    freq_array=freqs * units.GHz.to("Hz"),
+                    new_object=True,
+                    interpolation_function=pixel_interp,
+                    #                    freq_interp_func=freq_interp,
```
r-pascua, line 1097: Does the new API have a `freq_interp_func` parameter, or do we need to stick with what's on lines 1091-1092? If it's the latter, then let's remove this line; if it's the former, then let's remove lines 1091-1092 and uncomment this.

r-pascua (commented):
This looks fine, but I think we should deal with the weird commented-out bits before merging.

hera_sim/sigchain.py
steven-murray: yes.

hera_sim/sigchain.py
steven-murray: I don't believe it does, which was confusing to me.

r-pascua (approved):
This looks great! Thanks @steven-murray!

PR #268, opened 2023-08-08
Fix mutual coupling conjugation

r-pascua:
This PR fixes the conjugation error in the electric field propagator in the mutual coupling simulator. This should now ensure that the fringe-rate vs delay waterfalls for simulated coupling have the same shape as in the data.


steven-murray (approved):
This looks good, thanks @r-pascua. Figures in the notebook look great!

r-pascua:
This took a long time to come around to, but I've finally made a decisive comparison of the coupling simulations against what we see in the data. The amplitude of the coupling is greater in the data than what is predicted by the simulation, but it looks like the shapes in fringe-rate vs delay space agree. I figured I would post the results here for reference.

Screenshot from 2024-01-17 19-57-11

PR #269, opened 2023-08-14
Fix adjustment issues

r-pascua:
This fixes the broken tests for the `adjustment.py` module. The issues were due to phasing information (apparent RA, dec, frame position angle, and field IDs) now being a required attribute of `UVData` objects. Since these are "blt-like" arrays, they also need to be updated when interpolating the data to a different LST grid.


steven-murray (approved):
Looks good, thanks @r-pascua

PR #274, opened 2023-09-18
maint: update pypi publish action version

steven-murray:
Use the non-deprecated link to the pypi publish action

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -1,5 +1,8 @@
 name: Publish Python distributions to PyPI
 
+# The idea here is that you would make a Release via the web interface on github,
+# and this automatically then publishes to pypi
+
 on:
   push:
     tags:
@@ -9,9 +12,6 @@ jobs:
   build-n-publish:
     name: Build and publish to PyPI
     runs-on: ubuntu-latest
-    env:
-      ENV_NAME: publish
-      PYTHON: 3.8
     steps:
       - uses: actions/checkout@main
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/publish.yaml
```diff
@@ -20,7 +20,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v3
         with:
-          python-version: ${{ env.PYTHON }}
+          python-version: "3.11"
```
r-pascua, line 23: What's the motivation for hard-coding this instead of using the environment Python?

.github/workflows/publish.yaml
```diff
@@ -31,7 +31,7 @@ jobs:
 
       - name: Publish to PyPI
         if: startsWith(github.event.ref, 'refs/tags')
-        uses: pypa/gh-action-pypi-publish@master
+        uses: pypa/gh-action-pypi-publish@v1.8.8
```
r-pascua, line 34: Similar vein as the other question: why are we using a particular branch that's not the main branch?

r-pascua (commented):
I'd like to believe that this is a sensible and small change, but I get a little worried seeing specific versions of things pinned in code. Can you just quickly motivate why I shouldn't be worried about these changes?

.github/workflows/publish.yaml
steven-murray: The main/master branch has been "sunset" and is no longer receiving updates (according to the docs of the action). I can add a dependabot action that will auto-update this tag via PRs in the future.

.github/workflows/publish.yaml
r-pascua: gotcha, thanks for clearing that up!

.github/workflows/publish.yaml
r-pascua: gotcha

r-pascua (approved):
Thanks for the updates and answering my questions!

PR #291, opened 2023-12-28
feat: add option to not check autos

steven-murray:
Adds an option not to check autos at all, since this takes ~1min for a simulation that takes <2min.

### DIFF
--- a/hera_sim/visibilities/cli.py
+++ b/hera_sim/visibilities/cli.py
@@ -167,7 +167,7 @@ def run_vis_sim(args):
         comm.Barrier()
         sys.exit(0)
 
-    if myid == 0:
+    if args.run_auto_check:
         # Check imaginary of xx/yy autos and fix non-real values if the option is
         # selected in the arguments
         # xxpol = data_model.uvdata.get_data("xx")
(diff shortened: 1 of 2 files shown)


hera_sim/visibilities/cli.py
```diff
@@ -167,7 +167,7 @@ def run_vis_sim(args):
         comm.Barrier()
         sys.exit(0)
 
-    if myid == 0:
+    if args.run_auto_check:
```
piyanatk, line 170: The auto check should still be done by rank 0 process, right?

piyanatk (approved):
I think the auto checking and compression should be done by rank 0 process, right? Otherwise, look good, Steven.

hera_sim/visibilities/cli.py
steven-murray: Well, all other processes exit in the statement above, so only rank 0 does anything beyond that point. I thought it was nicer to not have the extra unnecessary indent.

piyanatk: Ah yes, I see now. The rank checking is not necessary then.

PR #306, opened 2024-04-18
Add fftvis simulator to hera_sim

tyler-a-cox:


hera_sim/visibilities/fftvis.py
```diff
@@ -0,0 +1,318 @@
...
+            + nt * 9  # rotation matrices
+            + 3 * nsrc
+            + 3 * nsrc  # source positions (topo and eq)
+        )
+
+        return all_floats * self._precision * 4 / 1024**3
+
+    def simulate(self, data_model):
+        """
+        Calls :func:fftvis to perform the visibility calculation.
```
steven-murray, line 220: After :func:, the fftvis should be surrounded in backticks

hera_sim/visibilities/fftvis.py
```diff
@@ -0,0 +1,318 @@
...
+            self._real_dtype = np.float32
+            self._complex_dtype = np.complex64
+        else:
+            self._real_dtype = float
+            self._complex_dtype = complex
+
+        if use_gpu:
+            raise ImportError("GPU acceleration is currently not supported for fftvis.")
+
+        self._fftvis = fftvis.simulate.simulate
```
steven-murray, line 93: This made sense in matvis because we had to switch between cpu/gpu, but here let's just use fftvis.simulate.simulate directly in the `simulate()` method

hera_sim/visibilities/fftvis.py
```diff
@@ -0,0 +1,318 @@
...
+
+        for i, freq in enumerate(data_model.freqs):
+            # Divide tasks between MPI workers if needed
+            if self.mpi_comm is not None and i % nproc != myid:
+                continue
+
+            logger.info(f"Simulating Frequency {i+1}/{len(data_model.freqs)}")
+
+            # Call fftvis function to simulate visibilities
+            vis = self._fftvis(
```
steven-murray, line 290: Repeating the above comment -- let's just use the `fftvis.simulate.simulate` directly here.

hera_sim/visibilities/fftvis.py
```diff
@@ -0,0 +1,318 @@
...
+                ants=active_antpos,
```
steven-murray, line 291: I think with fftvis that we can avoid some annoying difficulties here, by literally just passing in the baselines you want to simulate, instead of the antennas. FFTVis natually works directly on baselines, and this is easier to pass in (since they exist on the uvdata object already), easier after simulation since you don't need to do the `_reorder_vis` (which can actually be somewhat expensive), and probably easier within fftvis as well.

steven-murray (commented):
Thanks a bunch @tyler-a-cox, this is really very nice. I have a few comments, only one of which is really substantial.

hera_sim/visibilities/fftvis.py
```diff
@@ -0,0 +1,318 @@
...
+from .simulators import ModelData, VisibilitySimulator
+
+logger = logging.getLogger(__name__)
+
+
+class FFTVis(MatVis):
+    """
+    fftvis visibility simulator.
+
+    This is a fast, non-uniform FFT-based visibility simulator.
```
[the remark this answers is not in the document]
tyler-a-cox: Updated this

hera_sim/visibilities/fftvis.py
```diff
@@ -0,0 +1,318 @@
...
+        The "unpolarized" output of ``fftvis`` is expected to be XX polarization, which
+        corresponds to whatever the UVData object considers to be the x-direction
+        (default East).
+    precision : int, optional
+        Which precision level to use for floats and complex numbers.
+        Allowed values:
+        - 1: float32, complex64
+        - 2: float64, complex128
+    use_gpu : bool, optional
+        Whether to use the GPU version of fftvis or not. Default: False.
```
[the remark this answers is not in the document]
tyler-a-cox: Removed this. The developers of `finufft` think that there may be a GPU implementation of the transform we use for `fftvis` by the end of the summer, but for now I can just leave it out.

hera_sim/visibilities/fftvis.py
```diff
@@ -0,0 +1,318 @@
...
+    **kwargs
+        Passed through to :class:`~.simulators.VisibilitySimulator`.
+
+    """
+
+    conjugation_convention = "ant1<ant2"
+    time_ordering = "time"
+
+    diffuse_ability = False
+    # __version__ = __version__ # Fill in the version number here
```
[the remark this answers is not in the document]
tyler-a-cox: Updated this

hera_sim/visibilities/fftvis.py
```diff
@@ -0,0 +1,318 @@
...
+        self._precision = precision
+        if precision == 1:
+            self._real_dtype = np.float32
+            self._complex_dtype = np.complex64
+        else:
+            self._real_dtype = float
+            self._complex_dtype = complex
+
+        if use_gpu:
+            raise ImportError("GPU acceleration is currently not supported for fftvis.")
```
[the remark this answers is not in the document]
tyler-a-cox: Removed this

hera_sim/visibilities/fftvis.py
```diff
@@ -0,0 +1,318 @@
...
+    """
+
+    conjugation_convention = "ant1<ant2"
+    time_ordering = "time"
+
+    diffuse_ability = False
+    # __version__ = __version__ # Fill in the version number here
+
+    def __init__(
+        self,
```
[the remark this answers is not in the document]
tyler-a-cox: Updated this

hera_sim/visibilities/fftvis.py
```diff
@@ -0,0 +1,318 @@
...
+        self.kwargs = kwargs
+
+    def validate(self, data_model: ModelData):
+        """Checks for correct input format."""
+        # N(N-1)/2 unique cross-correlations + N autocorrelations.
+        # if data_model.uvdata.Nbls != data_model.n_ant * (data_model.n_ant + 1) / 2:
+        #    raise ValueError(
+        #        "FFTVis requires using every pair of antennas, "
+        #        "but the UVData object does not comply."
+        #    )
```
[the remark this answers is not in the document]
tyler-a-cox: Removed this

hera_sim/visibilities/fftvis.py
```diff
@@ -0,0 +1,318 @@
...
+
+        logger.info("Checking baseline-time axis shape")
+        if len(data_model.uvdata.data_array) != len(
+            data_model.uvdata.get_antpairs()
+        ) * len(data_model.lsts):
+            raise ValueError("FFTVis requires that every baseline uses the same LSTS.")
```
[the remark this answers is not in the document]
tyler-a-cox: Updated this in `FFTVis` and `MatVis`

hera_sim/visibilities/fftvis.py
```diff
@@ -0,0 +1,318 @@
...
+            # TODO: the following is extremely slow. If possible, it would be good to
+            # find a better way to do it.
+            if any(
+                len(data_model.uvdata.antpair2ind(ai, aj)) > 0
+                and len(data_model.uvdata.antpair2ind(aj, ai)) > 0
+                for ai, aj in data_model.uvdata.get_antpairs()
+                if ai != aj
+            ):
+                raise ValueError(
+                    "MatVis requires that baselines be in a conjugation in which "
```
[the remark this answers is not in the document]
tyler-a-cox: Updated this

hera_sim/visibilities/fftvis.py
tyler-a-cox: Fixed this

hera_sim/visibilities/fftvis.py
tyler-a-cox: Changed this to just call `fftvis.simulate.simulate`

hera_sim/visibilities/fftvis.py
tyler-a-cox: Fixed this

hera_sim/visibilities/fftvis.py
tyler-a-cox: I updated this to use to `uvdata.get_antpairs()`  to compute the baselines we need and changed the `FFTVis._reorder_vis` to be a little more efficient. It's currently passing all tests locally, but I may revisit this to see if I can make the reordering faster when `uvdata.blt_order != ("time", "baseline")`. Would be open to your thoughts if you can think of a more efficient way!

hera_sim/visibilities/fftvis.py
```diff
@@ -297,12 +290,13 @@ def simulate(self, data_model):
...
                 precision=self._precision,
                 polarized=polarized,
+                baselines=antpairs,
                 **self.kwargs,
             )[0]
 
             logger.info("... re-ordering visibilities...")
             self._reorder_vis(
-                req_pols, data_model.uvdata, visfull[:, 0, i], vis, ant_list, polarized
+                req_pols, data_model.uvdata, visfull[:, 0, i], vis, antpairs, polarized
```
steven-murray, line 299: Hmmm, is this "0" index a hold-over from the pre-future-array-shapes era? I think it is, and if so it should be removed...

hera_sim/visibilities/fftvis.py
```diff
@@ -316,3 +310,42 @@ def simulate(self, data_model):
             return 0
         else:
             return visfull
+
+    def _reorder_vis(self, req_pols, uvdata, visfull, vis, antpairs, polarized):
+        try:
+            if getattr(uvdata, "blt_order", None)[0] == "time":
```
steven-murray, line 316: this can now be written as `if not uvdata.time_axis_faster_than_bls:`

steven-murray: I think, actually, that this function can be a lot simpler... since `FFTVis` _only_ works for rectangular blts (you check this in the validation method), you should be able to do something like the following, for the whole function. This is because you're certain to have data coming out of FFTVis that has the same antpair ordering as the UVData (since you pass in the antpairs from the UVData to the simulation function)

```python
if uvdata.time_axis_faster_than_bls:
    vis = vis.transpose((3, 0, 1, 2))
else:
    vis = vis.transpose((0, 3, 1, 2))
 
if polarized:
    for p, (p1, p2) in enumerate(req_pols):
        visfull[:, p] = vis[:, p1, p2].reshape(-1)
    else:
        visfull[:, 0] = vis.reshape(-1)
```

Note that I think it would be even _better_, if possible, for the pols from the uvdata to be passed to the `simulate()` function, so the pols are automatically arranged in the same order as the uvdata, and also if possible for the antpairs axis in the `fftvis` output to come _before_ the pols (so you pass out something in the shape `(ntimes, nantpairs, npols)` OR `(ntimes, nantpairs, nfeed, nfeed)`).  That would require no reordering here. 

We can't do this in matvis because the structure of the outer product naturally puts the output into the form `(ntimes, nfeed, nfeed, nant, nant)`. But I think you could do this in `fftvis`.

steven-murray (commented):
Thanks @tyler-a-cox! Just the one comment -- I think we can still get this ordering slightly better / more intuitive.

hera_sim/visibilities/fftvis.py
tyler-a-cox: Yeah, this is due to the future-array-shapes problem. `MatVis` has a similar problem, which is why I chose to leave it this way. I'd be happy to `use_future_array_shapes` in both `FFTVis` and `MatVis`, but that will also require some rewriting of tests. I can either make those changes in this PR or open a separate PR once this is pushed through to make those changes.

hera_sim/visibilities/fftvis.py
tyler-a-cox: Thanks for the suggestion! This works great for most of the tests, but fails when `uvdata.reorder_blts(order='ant2')` is used. It seems like the reason for the failure came from my assumption that the order of `uvdata.get_antpairs()` reflected the baseline order in the data array, when that turned out not to be the case. I have a fix which gets the correct order of the baselines in the data and uses your solution. It may be a little hacky, so if there's a better way to can think to do this, I'd be happy to change it!

hera_sim/visibilities/fftvis.py
steven-murray: Ah I see.... yeah let's do this in a different PR. I'll make an issue :-)

hera_sim/visibilities/fftvis.py
```diff
@@ -245,8 +249,17 @@ def simulate(self, data_model):
...
+        # In certain cases, uvdata.get_antpairs() will return a different order
+        # than the order in the data_array
+        antpair_dict = {}
+        for ant1, ant2 in zip(
+            data_model.uvdata.ant_1_array, data_model.uvdata.ant_2_array
+        ):
+            if (ant1, ant2) not in antpair_dict:
+                antpair_dict[(ant1, ant2)] = None
+
+        antpairs = list(antpair_dict.keys())
```
steven-murray, line 262:
```suggestion
        # Get antpairs in the order that they appear in the uvdata.data_array
        # In certain cases, uvdata.get_antpairs() will return a different order
        # than the order in the data_array
        if data_model.uvdata.time_axis_faster_than_bls:
            antpairs = list(zip(data_model.uvdata.ant_1_array[::data_model.uvdata.Ntimes], data_model.uvdata.ant_2_array[::data_model.uvdata.Ntimes] ))
        else:
            antpairs = list(zip(data_model.uvdata.ant_1_array[:data_model.uvdata.Nbls], data_model.uvdata.ant_2_array[:data_model.uvdata.Nbls] ))
```

hera_sim/visibilities/fftvis.py
steven-murray: Got it -- I should have noticed that actually. Yeah `get_antpairs()` is not guaranteed to return antpairs in any given order. I made a suggestion (see below) which should be both faster and less hacky -- and in our case, always correct. Feel free to assign `data_model.uvdata` to some shorter variable name (e.g. `uvd`) higher up in the script to enhance readability...

hera_sim/visibilities/fftvis.py
tyler-a-cox: Sounds good! I'll work on that soon

steven-murray (approved):
This all looks good. I'm approving even though ubuntu tests are failing -- it's a problem with the mpi action, not the code.

steven-murray:
@tyler-a-cox sorry, I already approved, but I notice that we need an extra test or two, and some linting fixes.

PR #308, opened 2024-04-23
Update to new numpy RNG API

r-pascua:
This PR implements a shift to use the new API for generating random numbers with `numpy`. Now, rather than seed the global `numpy` random state prior to simulating a component, users should provide a `np.random.Generator` instance that is seeded as they desire. For folks who prefer to use the `Simulator`, then they effectively don't need to change anything&mdash;the interaction with the `Simulator` looks the same in basically every case (there are some corner cases, but, assuming I'm not missing something major, they're unusual enough that they should just be addressed in one-off conversations).

The following simple rule can be applied to update code to be compliant with this change:

**PRIOR TO THIS UPDATE**
```
seed = 12345  # or whatever you want to use
np.random.seed(seed)
cmp = sim_component(*args, **kwargs)
```

**AFTER THIS UPDATE**
```
seed = 12345  # or whatever you want to use
rng = np.random.default_rng(seed)
cmp = sim_component(*args, **kwargs, rng=rng)
```

In both cases, `args` are the positional arguments required by the simulation component's call signature, while `kwargs` are the keyword arguments that are typically stored as the class instance's attributes (i.e., these are the model parameters).

The most substantial change here is how the `Simulator` juggles random states. This is hidden from the end user, but it is worthwhile to note the changes in some detail here for those interested in how the `Simulator` internals function. Here is a high-level description of how the logic worked prior to this update, as well as a brief description of how it works after the changes made in this PR.

**PRIOR TO THIS UPDATE**

Before simulating the desired effect with the `add` method, the `Simulator` checks to see if the user has requested that the random state be managed in a particular way by checking the value of the `seed` parameter. If the user provided something, then the `Simulator` effectively enters a random state management routine before evaluating the model. Importantly, ***the random state is set during this random state management routine*** via the old API `np.random.seed(...)`. Seed caching is performed as-needed, depending on the type of seeding prescription requested.

**AFTER THIS UPDATE**

The initial logic is the same as before&mdash;the `Simulator` checks to see whether it should set the random state prior to evaluating the model. The main difference now is that the `_seed_rng` routine (the random state management routine mentioned above) returns both the (possibly updated) `seed` as well as the random number generator that has been seeded as desired. The `seed` returned is just to ensure that the `Simulator` has memory of what it was asked to do (which allows for cool stuff like making sure every baseline in a redundant group gets the same random seed, if desired), while ***the random number generator is passed along to the model and is ultimately used to make the random component of the model***.

### DIFF
--- a/hera_sim/foregrounds.py
+++ b/hera_sim/foregrounds.py
@@ -39,6 +39,8 @@ class DiffuseForeground(Foreground):
         Keyword arguments and associated values to be passed to
         :func:`~hera_sim.utils.rough_fringe_filter`. Default is to use the
         following settings: ``fringe_filter_type : tophat``.
+    rng: np.random.Generator, optional
+        Random number generator.
 
     Notes
     -----
@@ -76,6 +78,7 @@ class DiffuseForeground(Foreground):
         omega_p=None,
         delay_filter_kwargs=None,
         fringe_filter_kwargs=None,
+        rng=None,
     ):
         if delay_filter_kwargs is None:
             delay_filter_kwargs = {
@@ -91,6 +94,7 @@ class DiffuseForeground(Foreground):
             omega_p=omega_p,
             delay_filter_kwargs=delay_filter_kwargs,
             fringe_filter_kwargs=fringe_filter_kwargs,
+            rng=rng,
         )
 
     def __call__(self, lsts, freqs, bl_vec, **kwargs):
@@ -122,6 +126,7 @@ class DiffuseForeground(Foreground):
             omega_p,
             delay_filter_kwargs,
             fringe_filter_kwargs,
+            rng,
         ) = self._extract_kwarg_values(**kwargs)
 
         if Tsky_mdl is None:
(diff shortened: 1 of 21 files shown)


hera_sim/foregrounds.py
```diff
@@ -278,44 +289,44 @@ def __call__(self, lsts, freqs, bl_vec, **kwargs):
...
 
         # convolve vis with beam at each frequency
         for j, freq in enumerate(freqs):
-            # first calculate the beam, using truncated Gaussian model
+            # Treat the beam as if it's a Gaussian with a sharp horizon.
             beam = np.exp(-(has**2) / (2 * beam_width[j] ** 2))
             beam = np.where(np.abs(has) > np.pi / 2, 0, beam)
 
-            # who the hell knows what this does
+            # Compute the phase evolution as the source transits the sky.
```
steven-murray, line 324: this is a much better description, thank you :-)

steven-murray (approved):
Wow, great work @r-pascua, this looks really good.

PR #319, opened 2024-07-15
Update to use numpy=2 and pyuvdata=3

steven-murray:
This will probably not run OK until https://github.com/HERA-Team/hera_cal/pull/959 is merged.

This adds support for numpy2 and pyuvdata 3.0.

### DIFF
--- a/.github/workflows/test_suite.yaml
+++ b/.github/workflows/test_suite.yaml
@@ -19,7 +19,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest]
+        os: [ubuntu-latest, macos-13]
         python-version: ["3.10", "3.11", "3.12"]
 
     steps:
--- a/hera_sim/adjustment.py
+++ b/hera_sim/adjustment.py
@@ -313,6 +313,12 @@ def match_antennas(
                 for target_ant in target_copy.ant_2_array
             ]
         )
+        for i, bl in enumerate(target_copy.baseline_array):
+            ant1, ant2 =  target.baseline_to_antnums(bl)
+            ant1 = target_to_reference_map[ant1]
+            ant2 = target_to_reference_map[ant2]
+            newbl = target.antnums_to_baseline(ant1, ant2)
+            target_copy.baseline_array[i] = newbl
 
     attrs_to_update = tuple()
     if relabel_antennas:
@@ -359,7 +365,16 @@ def match_antennas(
             ]
         )
 
+    target_copy._clear_key2ind_cache(target_copy)
+    target_copy._clear_antpair2ind_cache(target_copy)
+
     # Now update the data... this will be a little messy.
+    # print("Target, from baselines: ", target.get_antpairs())
+    # print("Copy, from baselines: ", target_copy.get_antpairs())
+    # print("Target, from antarray: ", [x for x in np.unique([target.ant_1_array, target.ant_2_array], axis=1).T])
+    # print("Copy, from antarray: ", [x for x in np.unique([target_copy.ant_1_array, target_copy.ant_2_array], axis=1).T])
+    
+    # print(target_to_reference_map)
     for antpairpol, vis in target.antpairpol_iter():
         ant1, ant2, pol = antpairpol
         # Skip this visibility if we dropped one of the antennas.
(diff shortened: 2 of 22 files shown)


.github/workflows/test_suite.yaml
```diff
@@ -19,7 +19,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest]
+        os: [ubuntu-latest, macos-13]
```
steven-murray, line 22: This change should be temporary -- `fftvis` tests fail with the latest M1/M2 chips, but we don't know why yet. This restricts tests to the older intel chips. I'll make an issue and @tyler-a-cox will investigate, but I don't think it should hold us up.

hera_sim/adjustment.py
```diff
@@ -377,18 +386,25 @@ def match_antennas(
...
+
+        if blts is not None:
             # The new baseline has the same conjugation as the old one.
-            this_slice = (blts, slice(None), pol_inds[0])
+            this_slice = (
+                blts,
+                slice(None),
+                pol_inds[0].start,
+            )  # this_slice = (blts, slice(None), pol_inds[0].start)
+            # print("Got a slice off blts: ", this_slice)
```
r-pascua, line 399: Commented out code to be removed

hera_sim/adjustment.py
```diff
@@ -377,18 +386,25 @@ def match_antennas(
...
         else:  # pragma: no cover
             # The new baseline is conjugated relative to the old one.
             # Given the handling of the antenna relabeling, this might not actually
             # ever be called.
             this_slice = (conj_blts, slice(None), pol_inds[1])
             vis = vis.conj()
             new_antpairpol = new_antpairpol[:2][::-1] + (pol,)
-
+            # print("Got a slice off conj")
```
r-pascua, line 407: Another commented out print debug statement

hera_sim/adjustment.py
```diff
@@ -400,12 +416,14 @@ def match_antennas(
...
-            new_bl_int
-        )
+        # # Update the baseline array in case the antenna numbers got jumbled.
+        # # old_bl_int = target.antnums_to_baseline(ant1, ant2)
+        # # new_bl_int = target.antnums_to_baseline(*new_antpairpol[:2])
+        # # target_copy.baseline_array[target_copy.baseline_array == old_bl_int] = (
+        # #     new_bl_int
+        # # )
+        # target_copy._clear_key2ind_cache(target_copy)
+        # target_copy._clear_antpair2ind_cache(target_copy)
```
r-pascua, line 426: More commented out code to remove

hera_sim/defaults.py
```diff
@@ -274,8 +274,8 @@ def _check_config(self):
         """Check and warn if any keys in the configuration are repeated."""
         # initialize dictionaries that enumerate the key, value pairs
         # in the raw configuration dictionary
-        counts = {key: 0 for key in self().keys()}
-        values = {key: [] for key in self().keys()}
+        counts = dict.fromkeys(self().keys(), 0)
+        values = dict.fromkeys(self().keys(), [])
```
r-pascua, line 278: This is a bad idea. Using `dict.fromkeys(blah, mutable)` creates pointers to only one instance of `mutable`. For example, try running the following code:
```
mydict = dict.fromkeys([1,2,3], [])
mydict[1].append(3)
mydict[2].append(5)
print(mydict[3])
```

This is *not* an issue when using a dict comprehension (hence why it was done that way initially).

r-pascua (changes requested):
Thanks for tackling this! This all looks good, except for one issue that needs to be fixed (I left a comment explaining the issue and how to fix it). I figure we ought to add a test that will break in case this mistake is made again in the future, but I'll need some time to think about how to execute that test.

.github/workflows/test_suite.yaml
tyler-a-cox: I'll take a look at this soon. Thanks for making an issue!

hera_sim/defaults.py
steven-murray: Ah! I didn't realize this! Thanks for catching what would have been an awful bug.

steven-murray:
Thanks @r-pascua for the good review. I've addressed your points now!

hera_sim/adjustment.py
```diff
@@ -377,18 +386,23 @@ def match_antennas(
             new_antpairpol = antpairpol
 
         # Figure out how to slice through the new data array.
+        # print(new_antpairpol, target_copy.get_antpairs())
```
r-pascua, line 389: Commented out print debug statement

r-pascua (approved):
Looks good to me! Thanks @steven-murray!

PR #322, opened 2024-08-21
feat: ability to snap antpos to red grid

steven-murray:
Adds the ability to snap antenna positions to a grid before simulation.


tyler-a-cox (approved):
This looks good to me! Thanks @steven-murray!

PR #323, opened 2024-08-30
Use astropy coords matvis

steven-murray:
Modifes the interface to matvis to use the new astropy-coords branch. This makes things much simpler, actually.

### DIFF
--- a/hera_sim/visibilities/simulators.py
+++ b/hera_sim/visibilities/simulators.py
@@ -221,6 +221,11 @@ class ModelData:
         return self.uvdata.lst_array[unique_inds]
 
     @cached_property
+    def times(self):
+        """Times at which data is defined."""
+        return np.unique(self.uvdata.time_array)
+
+    @cached_property
     def freqs(self) -> np.ndarray:
         """Frequnecies at which data is defined."""
         return self.uvdata.freq_array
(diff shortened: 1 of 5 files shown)


hera_sim/visibilities/simulators.py
```diff
@@ -220,6 +220,11 @@ def lsts(self):
         _, unique_inds = np.unique(self.uvdata.time_array, return_index=True)
         return self.uvdata.lst_array[unique_inds]
 
+    @cached_property
+    def times(self):
+        """Times at which data is defined."""
+        return np.unique(self.uvdata.time_array)
```
piyanatk, line 226: Is `unique` still the best way to get unique times from pyuvdata? If we know the ordering and number of times and baselines, we can probably just slices the time_array. But this might be best implemented on pyuvdata.

setup.cfg
```diff
@@ -80,7 +78,7 @@ tests =
 vis =
     fftvis>=0.0.7
     line-profiler
-    matvis>=1.2.1
+    matvis@https://github.com/hera-team/matvis@use-astropy-coords
```
piyanatk, line 81: The PR for this branch is close and it has not been merged to main, I think.

piyanatk (approved):
Thanks for this @steven-murray . Just a comment on getting unique times. The astropy PR on matvis was close, and it looks like that branch has not been merged to main FYI.

piyanatk:
The tests are failing because it could not find `use-astropy-coords` branch of `matvis`. It is still there, so not sure what happening.

PR #332, opened 2024-10-11
maint: update API for new multi-processing fftvis

steven-murray:
This PR updates the FFTVis API so that it works with the multiprocessing-enabled fftvis.


tyler-a-cox:
Thanks for all the work on this, @steven-murray! I'm mostly okay with the changes here. I was wondering why the `PerturbedPolyBeam` in the `polybeam_simulation` [notebook](https://app.reviewnb.com/HERA-Team/hera_sim/pull/332/) has changed so significantly despite using the same coefficients? I'm less familiar with the `PolyBeam` class, so if these changes are expected, this comment can probably be ignored.

steven-murray:
@bhazelton yes you're exactly right -- the difference is due to me removing the unpolarized branch of code. I think I was equating "unpolarized" with "power" beam, and since I only wrote the efield beam evaluation, I removed the unpolarized option. I'm not the original author of the polybeam code, so I'm not totally sure I understand the intention behind having the polarized keyword here. Maybe @philbull can enlighten us further?

steven-murray:
@bhazelton and @tyler-a-cox I updated the `PolyBeam` to again take the `polarized` keyword (with default of False, like in previous versions), which restores the match. We can remove this at a later date as we see fit (I'll make an issue about it). Otherwise, everything should be running through fine, so let's get this merged!

hera_sim/beams.py
```diff
@@ -193,15 +195,56 @@ def _efield_eval(
...
+            interp_data = modulate_with_dipole(
+                az_grid, za_grid, f_grid, self.ref_freq, beam_values, fscale
+            )
+        else:
+            # Empty data array
+            interp_data = np.zeros(
+                (2, 2) + f_grid.shape, dtype=np.complex128
+            )
+            interp_data[1, 0, :, :] = beam_values  # (theta, n)
+            interp_data[0, 1, :, :] = beam_values  # (phi, e)
```
bhazelton: This has a very strange property that used to be present in some pyuvsim analytic beams that I think is extremely unphysical: The North feed only responds to zenith angle oriented polarized light and the East feed only responds to azimuthally-oriented polarized light. Are you sure this is what you want it to do?

I'd expect an unpolarized beam to respond equally to zenith-angle oriented and azimuthally oriented polarized light.

steven-murray: I'd opt to keep the behavior consistent with previous versions for now, and create an issue (we can tag @philbull into it for comments)

steven-murray: Actually, already did that: https://github.com/HERA-Team/hera_sim/issues/343

tyler-a-cox (approved):
Thanks for the PolyBeam fix @steven-murray, this looks good to me!

PR #346, opened 2025-01-10
maint: use pypi version of hera-calibration

steven-murray:
Updates to use the PyPI version of hera-calibration, which we forgot to fix in v4.3.0 :/


steven-murray:
@jsdillon I want to merge this so I can make a release so the whole stack is fixed up. I can do it, but tests are failing right now because I'm forcibly depending on a version of hera_cal that doesn't exist yet (>3.6.1). This is because v3.6.1 explicitly depends on numpy <2, while `hera_sim` and `matvis` now depend on `numpy>=2`. 

So, I'm happy to merge this as-is and trust that it will all work out, as long as we can make a `hera-cal` release pronto.

jsdillon:
That sounds good to me @steven-murray

PR #348, opened 2025-01-17
test: fix path in broken test

steven-murray:
Fixes a broken path in a test, and fixes a few warnings from accessing `uvd.x_orientation`.


bhazelton (approved):
Looks good to me!

PR #379, opened 2025-12-30
Fix persistent fftvis errors with macos

tyler-a-cox:


steven-murray:
@tyler-a-cox I think I have now fixed this problem (well, not really fixed it so much as just restricted our testing on macos-latest to python 3.13+). If you have time to fix the root problem with ray etc in fftvis, that would be great, and we can put the 3.11+3.12 macos tests back in, but I don't think it's as urgent now. I'll close this since all it does so far is what I already did in #386

tyler-a-cox:
Thanks @steven-murray! I’ll try to take another look soon, but it’s been difficult to track down. Hopefully I’ll find a fix soon

PR #387, opened 2026-04-16
Support for multibeam fftvis

steven-murray:
Depends on #386 

Adds support for multi-beam FFTVis, following https://github.com/tyler-a-cox/fftvis/pull/63


pyproject.toml
```diff
@@ -44,7 +44,7 @@ dependencies = [
 
 [project.optional-dependencies]
 vis = [
-  "fftvis @ git+https://github.com/tyler-a-cox/fftvis",
```
tyler-a-cox, line 47: Should we revert this back to just “fftvis”?

tyler-a-cox (changes requested):
This mostly looks good to me! I just had one quick question about the pyproject.toml

pyproject.toml
steven-murray: Yes!

steven-murray: Done

tyler-a-cox (approved):

PR #393, opened 2026-05-05
deps: add hera-cal and mpi4py to test requirements

steven-murray:
Ensures that hera-cal and mpi4py are installed when trying to run tests


.github/workflows/test_suite.yaml
```diff
@@ -59,3 +60,53 @@ jobs:
           fail_ci_if_error: true
         env:
           CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
+
+  tests-min-deps:
```
bhazelton, line 64: This is testing minimum versions but not testing without optional dependencies. Do you want to test without optional dependencies too?

bhazelton:
Looks like a lot of tests are failing, probably because you're passing a beam when it needs a beam list?

steven-murray:
@bhazelton this is now ready for review! 

The failing tests were fixed in a separate PR that updated the fftvis wrapper.

I think the failing codecov is fine in this case.

.github/workflows/test_suite.yaml
steven-murray: Yes! I have done that now.

bhazelton (approved):
Looks good to me!

Issue #349, opened 2025-01-20
Probably wrong thermal noise level for cross-pols


steven-murray:
We're seeing significantly low thermal noise on cross-pols in the HERA validation data. It might be that the cross-pol noise is getting assigned using a single autocorrelation instead of the geometric mean of both autos.

PR #397, opened 2026-05-19
Fix noise sim

r-pascua:
Addresses #349. Thought I had already resolved this in the past, but evidently not. This just adds a few lines of code necessary to ensure that noise simulated with the `Simulator` uses the correct amplitude in cross-polarized visibilities. The fix was just adding `autovis_i` and `autovis_j` as attributes to the `ThermalNoise` class, and updating the `attrs_to_pull` class attribute so that the `Simulator` knows to provide the autocorrelations for each antenna in the baseline.

A corresponding change was added to the logic in `Simulator._update_args` regarding how autocorrelation data is retrieved. It now assumes that the co-pol autocorrelations should be retrieved. For example, when retrieving the autos corresponding to `(0,1,'xy')`, the `Simulator` will now provide `(0,0,'xx')` for `autovis_i` and `(1,1,'yy')` for `autovis_j`. This might not be the correct thing to do for the old mutual coupling simulation code (which only considered autos-into-crosses coupling), but we don't use that model any longer (and we never bothered looking at the cross-pols for those simulations anyway).

A test has been added for the updated code in `ThermalNoise`, but not for the `Simulator`. I could probably cook something up for the `Simulator`, but it would take a while if we don't have somewhat realistic 4-pol test data saved.


steven-murray (approved):
Thank you @r-pascua!

Issue #414, opened 2026-06-30
Comparison with pyuvsim fails for multiple beams and polarized=False


steven-murray:
In the `test_compare_pyuvsim` test module, the case in which mutliple beams are used fails if `polarized=False`. It is strange because `polarized=False` works fine for each individual beam, it is only when they are combined in the same list that it fails.

PR #415, opened 2026-06-30
Fix fftvis beamidx

steven-murray:
Pass the `beam_idx` parameter through to FFTVis.

Also adds a test that compares against pyuvsim using multiple beams. Note that for some reason if `polarized=False` this test fails. I made https://github.com/HERA-Team/hera_sim/issues/414 to track this bug.


tyler-a-cox (approved):
This is looking good to me! We should probably make a note to fix the polarized=False case at some point though

PR #420, opened 2026-07-15
Precompute __antpair2ind_cache

r-pascua:
Validating the `ModelData` object can be needlessly expensive when setting up simulations at the scale of full HERA because of the check that only one of `(ai, aj)` or `(aj, ai)` show up for any antenna pair. This workaround can avoid the problem by precomputing the antenna pair to baseline-time index cache.

Unfortunately, after writing and testing this, I have come to the realization that we can probably speed up the code block in the validation section (e.g., [here](https://github.com/HERA-Team/hera_sim/blob/14a8080a58777ee4a25e68af67bb07a2e18c5767/src/hera_sim/visibilities/fftvis.py#L102)) with something like
```
antpairs = set(uvdata.get_antpairs())
if any((aj,ai) in antpairs for (ai,aj) in antpairs if ai != aj):
    raise ValueError(...)
```
instead of checking that either `uvdata.antpair2ind(ai,aj)` or `uvdata.antpair2ind(aj,ai)` returns `None`. The only corner case that I *think* I can see where this fails would be something like the `"nontrivial"` case in my tests that exclude setting the `blts_are_rectangular` attribute to `False` (while also having time as the fast part of the baseline-time axis). Although I think that case would fail silently, as it would ultimately tie back to `uvdata.get_antpairs()` not returning the full list of antenna pairs.

### DIFF
--- a/src/hera_sim/utils.py
+++ b/src/hera_sim/utils.py
@@ -751,6 +751,46 @@ def tanh_window(x, x_min=None, x_max=None, scale_low=1, scale_high=1):
     return window
 
 
+def precompute_antpair_index_cache(uvdata):
+    """Precompute the antpair2ind cache to reduce simulation setup costs.
+
+    The validation step performed during the initialization of a
+    ``visibilities.VisibilitySimulation`` object performs a check that the
+    baseline conjugation convention does not change with time. Under the hood,
+    this involves repeatedly calling ``UVData.antpair2ind``, which is very
+    inefficient for large N arrays---for the full HERA core, this can take
+    roughly 10 minutes. Precomputing the cache mapping antenna pairs to
+    blt-axis slices, which this function does, can drastically reduce that cost.
+    For non-rectangular baseline-times, the overall cost of precomputing the
+    cache and then validating the simulation parameters is expected to not be
+    reduced relative to validating the simulation without cache precomputation.
+
+    Parameters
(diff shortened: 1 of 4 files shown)


steven-murray:
Thanks @r-pascua --- yeah I think your solution in the OP is the easiest/best way to go. You're right that it would fail in that case, but that case is INVALID DATA anyway (unfortunately, it's possible to get into situations where data is invalid in pyuvdata -- this is due to the ability to update attributes in-place, which is always a tricky thing to manage).

r-pascua:
In that case, I figure I should just implement the suggested change in the OP and add some tests to ensure that it works for valid `UVData` configurations? I'm wondering about keeping the new function for precomputing the cache, since precomputing the cache should make `MatVis._reorder_vis` much faster in the case where we actually need to reorder the output from `matvis` (I'm guessing that the slow part in that function is also the repeated calls to `antpair2ind`). On a related note, I think we can use the lessons learned from this to also handle the `blts_are_rectangular and time_axis_faster_than_bls` case in `MatVis._reorder_vis`&mdash;should I add that feature as well?

steven-murray:
Yeah -- I reckon use your more simple example to fix this. If there's another place we can use the cache, let's use it there! The reorder_vis function is stupidly slow (when it's not short-circuited).

PR #427, opened 2026-08-02
fix: prevent xdist worker crashes from concurrent FFTVis JIT compiles

steven-murray:
## Summary

- Diagnoses the intermittent `Testing (ubuntu-latest, ...)` crash seen on #424 (Python 3.14) and again on a later run of this PR (Python 3.11): it's not a failing test, it's **all pytest-xdist workers dying at once** during collection (`node down: Not properly terminated`, pytest exit code 5, zero tests ever collected — an OOM-kill signature, no traceback).
- `-n auto` resolves to just **2** workers on these runners (`psutil`-detected physical CPU count) — the `gw0`...`gw9` worker IDs in the crash log are xdist's restart attempts on those 2 slots (default `--max-worker-restart` = `numprocesses * 4` = 8), not 10 concurrent processes. So this isn't primarily about many workers racing on one heavy test; it's the controller *and* each worker independently importing the full heavy scientific stack (jax, numba, healpy, scipy, matvis, fftvis, ...) concurrently, on a memory-constrained 2-core runner, occasionally tipping into an OOM kill during collection. Recurrence on 3.11 (no special numba build there) rules out a Python-3.14-specific cause.
- `tests/test_visibilities/test_fftvis.py::test_snap_positions` is still real waste worth trimming — it runs three full `FFTVis().simulate()` calls and is 4-9x slower than any other test (83–93s locally) — so the FFTVis-grouping fix stays in as a genuine improvement, just not the whole fix.

## Changes

1. Group all FFTVis-backed tests onto a single xdist worker (`xdist_group` marker via a `pytest_collection_modifyitems` hook, plus `--dist=loadgroup`), so only one process pays FFTVis's numba JIT-compile cost.
2. Install only the `tests` dependency group (not the full `dev` group, which also pulls in the entire Sphinx/docs toolchain) for the two ubuntu-latest jobs that run the full suite with `-n auto`, cutting unnecessary install/disk footprint.
3. Retry the pytest invocation up to 3 times, but only when it exits with code 5 (pytest's "no tests collected" code, matching this crash signature exactly) — genuine test failures (exit 1) still fail the job immediately without masking anything.

## Test plan

- [x] Verified locally on Python 3.14 (matching PR #424's dependency versions) that FFTVis tests are now confined to a single `gw` worker under `-n 4 --dist=loadgroup`.
- [x] Full test suite passes locally: 609 passed, 3 xfailed.
- [x] Verified the retry-loop shell logic in isolation for the three cases (success, exit-5-then-success, exit-5-exhausted, non-5-failure).
- [ ] Confirm CI is stable across repeated runs (the underlying issue is a resource-pressure race, so absence of a crash on one run isn't proof by itself — the retry is the actual safety net).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


tyler-a-cox (approved):
This all looks good to me. Thanks @steven-murray!

PR #439, opened 2026-09-09
Fix flaky CI worker crashes via NUMBA_CPU_NAME=generic

steven-murray:
## Summary

Both [#435](https://github.com/HERA-Team/hera_sim/pull/435) and [#433](https://github.com/HERA-Team/hera_sim/pull/433) were failing the `Tests (upgraded deps)` check, apparently at random. It isn't actually flaky test logic — it's numba crashing during pytest collection.

numba's `config.py` calls `llvmlite.binding.get_host_cpu_name()` **eagerly at import time** (to pick a default for `NUMBA_ENABLE_AVX`). Since `hera_sim` imports numba at module load (`src/hera_sim/utils.py`), just importing `hera_sim` during test collection triggers this call. Some GitHub-hosted runners have CPUs that misreport their AVX/AVX-512 support to virtualized guests, which crashes that call outright — numba's own source has a comment acknowledging this exact class of bug ("virtual machines may incorrectly report AVX support", referencing numba/numba#9582).

This explains the observed symptoms:
- Crash happens during collection, before any test runs (`no tests ran in ~13s`)
- 100% reproducible across all 3 retry attempts in the existing retry loop, not randomly transient
- Only `Tests (upgraded deps)` was hit, since it's the job guaranteed to resolve the newest numba/llvmlite each run
- I reproduced the exact resolved dependency set locally (609 tests, all extras) under matching CPU/memory constraints and the full suite passed cleanly — ruling out the "OOM-killed workers" theory the existing retry-loop comment in the `tests` job assumed

## Fix

Set `NUMBA_CPU_NAME=generic` in the job env, which short-circuits the crashing call (`CPU_NAME or ll.get_host_cpu_name()` in numba's source) by skipping the host-CPU probe entirely. Verified this doesn't affect correctness: `test_mutual_coupling` and `test_reshape_vis` (which exercise hera_sim's numba-accelerated code paths) pass identically with the var set.

Applied to both `tests-upgraded` (the job that's actually failing) and the main `tests` matrix job, since that job's retry loop guards against the identical crash signature and was likely misdiagnosing the same numba issue as OOM.

## Test plan

- [x] Reproduced the exact resolved dependency set (numba 0.67.0, llvmlite 0.49.0, Python 3.14) locally and confirmed the full test suite passes
- [x] Confirmed `NUMBA_CPU_NAME=generic` doesn't break numba JIT compilation/execution locally
- [x] Ran hera_sim's numba-accelerated tests (`test_mutual_coupling`, `test_reshape_vis`) with the var set — all pass
- [ ] Confirm `Tests (upgraded deps)` passes on this PR's own CI run

🤖 Generated with [Claude Code](https://claude.com/claude-code)


tyler-a-cox (approved):
Okay, this looks good to me. Thanks @steven-murray!

